PR #3, opened 2022-07-11
Initial code for plugin based on opensearch-plugin-template-java

kevinawskendra:
### Description
Initial code for plugin based on opensearch-plugin-template-java

- Copied files from https://github.com/opensearch-project/opensearch-plugin-template-java (excluding `CONTRIBUTING.md`, `LICENSE.txt`, `NOTICE.txt`, and `README.md`) then followed README.

- Renamed `LICENSE` and `NOTICE` to `LICENSE.txt`, `NOTICE.txt`, respectively, in order for `./gradlew check` to succeed.

### Issues Resolved
N/A

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


ps48:
@kevinawskendra Thanks for the PR. Below are some recommended changes: 
1. Please use [commit signoffs](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff) otherwise the DCO workflow would fail. 
2. Before adding the plugin template code. Let's add an RFC design document created for this new plugin. You may checkout an [example here](https://github.com/opensearch-project/search-relevance/issues/2).

macohen:
Mahita, we can review this request after you resolve the DCO issue in your commits.

semantic-reranker/src/main/java/org/opensearch/search/relevance/preprocess/SlidingWindowTextSplitter.java
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * The minimum step size by which to move the sliding window after a split. Step size can be larger
+   * in practice because of respecting sentence boundaries.
+   */
+  private Integer stepSize;
+
+  /**
+   * The maximum number of passages to be extracted from the input text
+   */
+  private Integer maximumPassages;
```
msfroh: Why are these `Integer` rather than `int`?

The contract of `setSlidingWindow` implicitly says that they can't be `null`.

semantic-reranker/src/main/java/org/opensearch/search/relevance/client/KendraHttpClient.java
```diff
@@ -0,0 +1,93 @@
...
+      try {
+        rerankRequest.setRerankingEndpointId(endpointId);
+        System.out.println(rerankRequest.getRerankingEndpointId());
+        System.out.println(rerankRequest.getDocuments());
+        System.out.println(rerankRequest.getQueryText());
+        Request<Void> request = new DefaultRequest<>(aws4Signer.getServiceName());
+        request.setHttpMethod(HttpMethodName.POST);
+        request.setEndpoint(URI.create(serviceEndpoint));
+        request.setHeaders(Map.of("Content-Type", "application/x-amz-json-1.0", "X-Amz-Target", "AWSKendraRerankingFrontendService.Rerank"));
+        System.out.println("objectMapper: " + objectMapper.writeValueAsString(rerankRequest));
```
msfroh: We'll need to remove these `System.out.println` before merging.

semantic-reranker/src/main/java/org/opensearch/search/relevance/client/KendraHttpClient.java
```diff
@@ -0,0 +1,93 @@
...
+        request.setContent(new ByteArrayInputStream(objectMapper.writeValueAsString(rerankRequest).getBytes(StandardCharsets.UTF_8)));
```
msfroh: There will be a dedicated Kendra reranking client in the AWS SDK, right?

I'm assuming this code (and the whole `dto` package) are temporary until you have added the client to the SDK.

semantic-reranker/src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
```diff
@@ -0,0 +1,370 @@
...
+          searchResponse.writeTo(out);
+
+          final StreamInput in = new NamedWriteableAwareStreamInput(out.bytes().streamInput(), namedWriteableRegistry);
+
+          LOGGER.info("Extracting search hits");
+          final SearchHits hits = new SearchHits(in);
+
+          final SearchHits newHits = doSemanticRerank(queryParserResult, hits, pluginEnabledFetchSource);
+
+          LOGGER.info("Extracting remaining response fields");
```
msfroh: I feel like this section can (and should) be extracted to support any external reranker. Anyone who is trying to manipulate the `SearchHits` of a `SearchResponse` would need to jump through these same hoops.

More generally, how hard would it be to split out the generic reranking logic from the Kendra-specific implementation? 

I think the generic logic is probably:
1. Call an abstract `shouldRerank` method based on the search request, and if true, create a generic response listener with a `Reranker`.
2. The generic response listener extracts the `SearchHits` from the response, passes them to the reranker, gets back new `SearchHits` and uses them (and the old `SearchResponse`) to create a new `SearchResponse`.

IMO, that would also help for breaking up this class (which is a little big right now).

msfroh (changes requested):

semantic-reranker/src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
```diff
@@ -0,0 +1,370 @@
...
+        SearchHit searchHit = idToSearchHitMap.get(rerankResultItem.getDocumentId());
+        if (searchHit == null) {
+          LOGGER.warn("Response from external service references hit id {}, which does not exist in original results. Skipping.",
+              rerankResultItem.getDocumentId());
+          continue;
+        }
+        if (pluginEnabledFetchSource) {
+          // User disabled fetching document source, so remove from response
+          searchHit.sourceRef(null);
+        }
```
msfroh: It took me a while to understand that `pluginEnabledFetchSource` meant "The user did not request the document source, the plugin did."

Maybe name this `suppressSourceOnResponse` or something and comment when you initialize it?

semantic-reranker/src/main/java/org/opensearch/search/relevance/model/dto/Document.java
```diff
@@ -0,0 +1,82 @@
...
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+import java.util.List;
+
+@JsonNaming(PropertyNamingStrategies.UpperCamelCaseStrategy.class)
+public class Document {
+  
+  private String id;
+  private String title;
```
msfroh: The existence of the `title` field looks like it's just part of the Kendra reranking API, and this plugin always passes `null`. 

The only fields that matter for this plugin are `id`, `tokenizedBodyPassages`, and `originalScore`, right?

(I'm also assuming that these classes will be replaced with a proper SDK client model.)

semantic-reranker/src/main/java/org/opensearch/search/relevance/preprocess/TextTokenizer.java
```diff
@@ -0,0 +1,49 @@
...
+          "can", "will", "just", "don", "should", "now"));
+
+  public List<List<String>> tokenize(List<String> texts) {
+    return texts.stream()
+        .map(text -> tokenize(text))
+        .collect(Collectors.toList());
+  }
+
+  public List<String> tokenize(String text) {
+    String[] tokens = text.split("[\\p{Punct}\\s]+");
```
msfroh: Written like this, the tokenizer will compile the regex on every invocation.

You should use define a constant `Pattern` field, initialized with `Pattern.compile` to capture the regex, and use the `Pattern`'s `split` method to split strings.

msfroh (changes requested):

macohen:
When this PR is complete, will we be able to close out https://github.com/opensearch-project/search-relevance/issues/4?

semantic-reranker/src/main/java/org/opensearch/search/relevance/preprocess/SlidingWindowTextSplitter.java
mahitamahesh: Updated to `int`

semantic-reranker/src/main/java/org/opensearch/search/relevance/client/KendraHttpClient.java
mahitamahesh: Left by accident. Removed now.

mahitamahesh:
@macohen Yes, I think so. We are looking to make some updates to the document pre-processing logic in a follow up PR, but that won't contradict anything mentioned in #4.

semantic-reranker/src/main/java/org/opensearch/search/relevance/client/KendraHttpClient.java
mahitamahesh: Yes, the service is currently under development so we don't have an external SDK. We will update this code once it is ready.

semantic-reranker/src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
mahitamahesh: Thanks for the suggestion, I've refactored in the latest commit.

semantic-reranker/src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
mahitamahesh: Thanks for the naming suggestion, I've updated in the latest commit and tried to re-word the comment to make the purpose clearer. Let me know what you think.

semantic-reranker/src/main/java/org/opensearch/search/relevance/preprocess/TextTokenizer.java
mahitamahesh: Thank you for catching this. Updated.

semantic-reranker/src/main/java/org/opensearch/search/relevance/model/dto/Document.java
mahitamahesh: We are still figuring out the user experience for configuring a mapping between fields in their document and the `title` and `body` expected by the ranker, so I will address this in a follow up.

semantic-reranker/src/main/java/org/opensearch/search/relevance/control/KendraSearchExtBuilder.java
```diff
@@ -17,29 +18,35 @@
...
+import org.opensearch.search.relevance.constants.Constants;
 
 public class KendraSearchExtBuilder extends SearchExtBuilder {
-  public static final String NAME = "kendra_intelligent_ranking";
+  public static final String NAME = Constants.PLUGIN_NAME;
 
   private static final ObjectParser<KendraSearchExtBuilder, Void> PARSER;
-  private static final ParseField RANKER_ENABLED = new ParseField("enabled");
+  private static final ParseField RANKER_ENABLED = new ParseField(Constants.ENABLED);
+  private static final ParseField BODY_FIELD = new ParseField(Constants.BODY_FIELD);
```
msfroh: I was chatting with some folks about the long-term strategy where we'll support multiple rerankers, and trying to make sure that users of this plugin will be able to send the same queries in the future.

The proposed spec is described in https://github.com/opensearch-project/search-relevance/issues/12.

Could we change the ext parameters to something more like:

```
{
  "ext": {
    "rankers": [
      {
        "name" : "kendra_intelligent_ranking",
        "properties" : {
          "body_field": [
            "field1",
            "field2"
          ]
        }
      }
    ]
   }
 }

Comment (mahitamahesh) [reply semantic-reranker/src/main/java/org/opensearch/search/relevance/control/KendraSearchExtBuilder.java]: I want to propose a slightly modified structure
```
{
  "ext": {
    "relevancy_config": {
      "rankers": [
        {
          "name" : "kendra_intelligent_ranking",
          "enabled": true
          "properties" : {
            "body_field": [
              "field1",
              "field2"
            ]
          }
        }
      ]
    }
  }
}
```

1. We need an ExtBuilder NAME to identify the ext config specified by this plugin, under which the inputs will be nested (I’m using `relevancy_config` used as a placeholder). See corresponding [code][1] and [usage][2] from LTR as an example.
2. We need to have a flag to enable/disable rankers if we are to support index-level and request-level configuration. For example:
```
PUT my-index-000001/_settings
{
  "index" : {
    "kendra_intelligent_ranking" : {
        "enabled" : "true",
        "body_field": "text"
    }
  }
}
```
will enable the Kendra ranker for `my-index-000001`, but issuing a query without an `ext` entry like below, will override this and disable the Kendra ranker for this request, when in fact, it should be enabled.
```
GET my-index-000001/_search
{
  "query": {
    "match": {
      "text": {
        "query": "research"
      }
    }
  }
}
```

[1]:https://github.com/o19s/elasticsearch-learning-to-rank/blob/9f0612cf24fe5d700cc4ab6fa274c09d51841d91/src/main/java/com/o19s/es/ltr/logging/LoggingSearchExtBuilder.java#L38

[2]: https://elasticsearch-learning-to-rank.readthedocs.io/en/latest/logging-features.html#logging-values-for-a-live-feature-set

**Comment by mahitamahesh** [comment]
On another note, we have implemented plugin controls via index settings, whereas #12 makes use of the `_meta` field under `my-index-000001/_mapping`

1. Has anyone built a proof of concept with `_meta` to test that the integration works for this use case? If we need to make the change from settings to mapping, can we base our implementation off of this PoC?
2. If we support purely UI-based configuration in the future, we can make the API change from settings to mapping under the hood and not break customer workflows. Else I think this would be a breaking change.

**Comment by mahitamahesh** [comment]
@msfroh Summarizing our offline discussion here:

### I. Transformer Interfaces
The plugin will invoke specific transformers at different points during the query flow (e.g. rewriters have to be applied upfront, whereas rankers will be invoked just before response, once all documents have been fetched). Further, the interface for each transformer type will differ based on the parameters it operates upon (e.g. rewriters operate on the query, rankers operate on `SearchHits`).
 
**Launch proposal:**
- Support one transformer type: `result_transformer` (`request_transformer`, etc. to be added in the future)
- Have a generic `ResultTransformer` interface with the methods below, and with a concrete implementation for `KendraIntelligentRanking`.
  - ```boolean shouldApplyTransformer (SearchRequest)```
  - ```SearchHits apply (SearchHits)```
- Modify plugin logic to invoke a list of `ResultTransformers` (we will launch with a list of size 1), each of which operates on the `SearchHits` from the previous stage
- We will not support passing context from one transformer to the next. This can be added in the future.
 
 
### II. Search Configuration (for Index and Request level)
Support index-level configuration via index settings (not `_mapping`, as proposed in #12 
Support request-level configuration via `ext` field in the request DSL.
 
Sample configuration based on #12 is at the end of this comment.
 
**Launch proposal:**
- Maintain a list of supported transformer types (currently `["result_transformer"]`) to perform validation
- Have a generic transformer configuration with attributes below. Each implemented transformer will define its properties and logic for parsing + validating them
  - `name`
  - `properties`
- Add unit tests with dummy operations to test that a chain of transformers works


#### Sample Configuration
```
"search_configuration": {
  "request_transformers": [
    {
      "name": "querqy",
      "properties": {
        "querqy": {
          "matching_query": {
            "must_match": {
              "query": "rambo"
            },
            "multi_match": {
              "query": "rambo",
              "fields": [
                "field1",
                "field2"
              ]
            }
          },
          "query_fields": [
            "title^3.0",
            "brand^2.1",
            "shortSummary"
          ]
        }
      }
    }
  ],
  "result_transformers": [
    {
      "name": "kendra_intelligent_ranking",
      "properties": {
        "title_fields": [
          "title"
        ],
        "body_fields": [
          "published",
          "description"
        ]
      }
    },
    {
      "name": "xyz_highlighting_service",
      "properties": {
        "prop1": "val1"
      }
    }
  ]
}

```

**Comment by msfroh** [comment]
Thanks @mahitamahesh ! I really like the API as described above. 

I think it satisfies the needs of the Kendra-based reranker calls, but will also make it easy to integrate with other rerankers (and other result transformers) soon.

**Comment by mahitamahesh** [comment]
Minor tweaks compared to the proposal [above](https://github.com/opensearch-project/search-relevance/pull/3#issuecomment-1279597651):

### I. Result Transformer Interface

```
List<Setting<?>> getTransformerSettings();

boolean shouldTransform(final SearchRequest request, final ResultTransformerConfiguration configuration);

SearchHits transform(final SearchHits hits, final SearchRequest request, final ResultTransformerConfiguration configuration);
```

### II. Configuration

#### Index settings
-  Index settings does not support lists of objects `Error: only value lists are allowed in serialized settings`
- Naming following [OpenSearch plugin conventions](https://github.com/opensearch-project/opensearch-plugins/blob/main/CONVENTIONS.md#settings), hence `index.plugin.searchrelevance.*`

```
PUT my-index/_settings

{
  "index": {
    "plugin": {
      "searchrelevance": {
        "result_transformers": {
          "kendra_intelligent_ranking": {
            "order": 2,
            "properties": {
              "title_fields": [
                "title"
              ],
              "body_fields": [
                "published",
                "description"
              ]
            }
          },
          "transformer2": {
            "order": 1,
            "properties": {
              "prop1": "val1"
            }
          }
        }
      }
    }
  }
}
```

#### Request-level configuration

- Moved `result_transformer` directly under `ext`
- We will add a new `SearchExtBuilder` class for each transformer type that is added

```
GET my-index/_search
{
  "query": {
    "match": {
      "text": {
        "query": "research"
      }
    }
  },
  "ext": {
    "result_transformer" : {
      "kendra_intelligent_ranking": {
        "order": 2,
        "properties": {
          "title_fields": [
            "title"
          ],
          "body_fields": [
            "published",
            "description"
          ]
        }
      },
      "transformer2": {
        "order": 1,
        "properties": {
          "prop1": "val1"
        }
      }
    }
  }
}
```

**Comment by msfroh** [comment]
> Moved `result_transformer` directly under `ext`

What was the motivation for that change? (I don't think it's an irreversible change, but I'm curious about why it was needed.)

File: semantic-reranker/src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/KendraIntelligentRanker.java [hunk 9]
```diff
@@ -0,0 +1,178 @@
...
+  private static final double BM25_B_VALUE = 0.75;
+  private static final double BM25_K1_VALUE = 1.6;
+  private static final int TOP_K_PASSAGES = 3;
+
+  private static final Logger logger = LogManager.getLogger(KendraIntelligentRanker.class);
+
+  private final OpenSearchClient openSearchClient;
+  private final KendraHttpClient kendraClient;
+  private final TextTokenizer textTokenizer;
+  private final CloseableHttpClient httpClient;
```
msfroh: This field is no longer used, and can be removed.

I though that we might also be able to remove the Apache HttpClient/HttpCore dependencies, but they need to be included so the AwsHttpClient still works.

semantic-reranker/src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/configuration/KendraIntelligentRankerSettings.java
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * The secret key (ie password) for connecting to Kendra.
+   */
+  public static final Setting<SecureString> SECRET_KEY_SETTING = SecureSetting.secureString("kendra_intelligent_ranking.aws.secret_key", null);
+
+  /**
+   * The session token for connecting to Kendra.
+   */
+  public static final Setting<SecureString> SESSION_TOKEN_SETTING = SecureSetting.secureString("kendra_intelligent_ranking.aws.session_token", null);
```
msfroh, line 71: Does it ever make sense to specify a session token in the settings?

AFAIK, session tokens are always short-lived, so it never makes sense to include one in the opensearch.yml.

semantic-reranker/src/main/plugin-metadata/plugin-security.policy
```diff
@@ -0,0 +1,16 @@
...
+ * compatible open source license.
+ */
+ 
+grant {
+  permission java.lang.RuntimePermission "accessDeclaredMembers";
+  permission java.lang.reflect.ReflectPermission "suppressAccessChecks";
+  
+  permission java.net.SocketPermission "*", "connect,resolve";
+  permission java.lang.RuntimePermission "getClassLoader";
+  permission java.util.PropertyPermission "*", "read,write";
```
msfroh: Can this permission be scoped down? 

In particular, I think we can limit it to `kendra_intelligent_ranking.*` properties.

semantic-reranker/config/external_service.yml
```diff
@@ -0,0 +1 @@
+# possible location for external configs
```
msfroh: Can we delete this file?

semantic-reranker/build.gradle
```diff
@@ -0,0 +1,133 @@
...
+    implementation 'org.apache.commons:commons-lang3:3.12.0'
+    implementation 'org.apache.httpcomponents:httpclient:4.5.13'
+    implementation 'org.apache.httpcomponents:httpcore:4.4.12'
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.13.2'
+    implementation 'com.fasterxml.jackson.core:jackson-core:2.13.2'
+    implementation 'com.fasterxml.jackson.core:jackson-annotations:2.13.2'
+    implementation 'commons-logging:commons-logging:1.2'
+    implementation 'com.amazonaws:aws-java-sdk-sts:1.12.300'
+    implementation 'com.amazonaws:aws-java-sdk-core:1.12.300'
+    implementation 'joda-time:joda-time:2.10.12'
```
msfroh: Looks like we can remove the dependency on `joda-time`.

semantic-reranker/build.gradle
```diff
@@ -0,0 +1,133 @@
...
+
+repositories {
+    mavenLocal()
+    maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
+    mavenCentral()
+    maven { url "https://plugins.gradle.org/m2/" }
+}
+
+dependencies {
+    implementation 'com.google.guava:guava:21.0'
```
msfroh: The uses of Guava in SlidingWindowTextSplitter and SearchRelevancePlugin seem trivial enough to implement with built-in Java classes, so I would drop the Guava dependency.

The replacements needed are:
```
Lists.newArrayList()     ->    new ArrayList<>()
ImmutableList.of(...)    ->    Arrays.asList(...)
```

msfroh (commented):
Adding a few nitpicky comments

msfroh:
@mahitamahesh -- how would you feel about us merging this PR, and doing cleanup in subsequent PRs?

IMO, we're functionally complete and the general code structure is good. We have the cleanup items I listed above (which I'm happy to do) and maybe some minor tweaks here and there, but I think piling more on to this PR will make it harder to follow. What do you think?

mahitamahesh:
@msfroh I have a commit ready addressing all of the cleanup items above, so I'll squeeze that in here if that's alright, and then we can call this PR done.

semantic-reranker/src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/KendraIntelligentRanker.java
mahitamahesh: Removed

semantic-reranker/src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/configuration/KendraIntelligentRankerSettings.java
mahitamahesh: Leaving this here since using session key is completely optional. People could use it for testing if long lived credentials are not set up yet.

semantic-reranker/src/main/plugin-metadata/plugin-security.policy
mahitamahesh: Removed entirely

semantic-reranker/config/external_service.yml
mahitamahesh: Done

semantic-reranker/build.gradle
mahitamahesh: Done

semantic-reranker/build.gradle
mahitamahesh: Removed Guava. Will likely remove lang3 in a follow up PR too.

msfroh (approved):

kevinawskendra:
LGTM

PR #22, opened 2022-11-09
Change plugin name and depend on 2.4

msfroh:
As we get closer to releasing this plugin to work with 2.4, we should update the name -- going with search-processor -- to better reflect the idea that it will be able to transform search requests and responses. Also, we need to update the build to point to 2.4 snapshots, instead of 2.1.

### Description
* Changes the name of the plugin artifact from `search-relevance` to `search-processor`. (We still need to change the name of the classes, but that's an internal detail, whereas this is the public-facing name of the plugin.)
 * Updates to build against OpenSearch-2.4.0-SNAPSHOT

This commit adds no new functionality.

### Issues Resolved
N/A
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ apply plugin: 'opensearch.pluginzip'
 
 group = 'org.opensearch'
 
-def pluginName = 'search-relevance'
+def pluginName = 'search-processor'
 def pluginDescription = 'Make Opensearch results more relevant.'
 def projectPath = 'org.opensearch'
 def pathToPlugin = 'search.relevance'
@@ -52,7 +52,7 @@ validateNebulaPom.enabled = false
 
 buildscript {
     ext {
-        opensearch_version = "2.1.0-SNAPSHOT"
+        opensearch_version = "2.4.0-SNAPSHOT"
     }
 
     repositories {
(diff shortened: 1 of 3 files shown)


msfroh:
Note that I haven't updated the name used in the index settings API.

For example, if you want to enable Kendra ranking by default for an index, the payload still looks like:

```json
PUT <index_name>/_settings
{
  "index": {
    "plugin" : {
      "searchrelevance" : {
        "result_transformer" : {
          "kendra_intelligent_ranking": {
              "order": 1,
              "properties": {
                "title_field": "title",
                "body_field": "text"
              }
          }
        }
      }
    }
  }
}
```

My thinking is that the index settings API is very likely to change in future anyway if we start supporting stored search configurations.

My preference for that future API would be something more like:

```json
PUT /_search_configuration/with_kendra_ranking
{
  "result_transformer": [
    {
      "name": "kendra_intelligent_ranking",
      "properties":  {
        "title_field": "title",
        "body_field": "text"
      }
    }
  ]
}

PUT /_search_configurations/with_other_ranking
{
  "result_transformer": [
    {
      "name": "some_other_ranker", // e.g. Metarank
      "properties":  {
        // component-specific properties
      }
    }
  ]
}

PUT <index_name>/_settings
{
  "index": {
    "plugin": {
      "search_processor": {
        "default_search_configuration": "with_other_ranking"
      }
    }
  }
}
```

mahitamahesh:
You can update the name used in index settings by updating [this PLUGIN_NAME constant](https://github.com/opensearch-project/search-relevance/blob/main/src/main/java/org/opensearch/search/relevance/configuration/Constants.java#L13)

msfroh:
> You can update the name used in index settings by updating [this PLUGIN_NAME constant](https://github.com/opensearch-project/search-relevance/blob/main/src/main/java/org/opensearch/search/relevance/configuration/Constants.java#L13)

Right -- I'm consciously avoiding changing that, because I don't want to break all of the work-in-progress documentation.

IMO, we will (eventually) make a breaking change to the index settings, and we can change the name then. Fortunately, it's more of a one-off to fix it. (Also, I firmly believe that most users will prefer the search request `ext` approach to enabling reranking until we add named search configurations.)

mahitamahesh:
Makes sense to me!

mahitamahesh (approved):

build.gradle
```diff
@@ -128,4 +128,4 @@ sourceSets {
 // TODO: Enable these checks
```
macohen, line 128: When do we need to enable these checks and fulfill the TODO?

macohen (approved):
Enabling the checks on line 129 or removing the TODO should be another task.

PR #23, opened 2022-11-10
Bug fix: Fix NPE when parsing field names in query parser.

mahitamahesh:
### Description
Bug fix: Fix NPE when parsing field names in query parser.
 
### Issues Resolved
No existing issue
 
### Check List
- [Y ] New functionality includes testing.
  - [Y ] All tests pass
- [N/A ] New functionality has been documented
  - [N/A ] New functionality has javadoc added
- [Y ] Commits are signed as per the DCO using --signoff 

### DIFF
--- a/src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/preprocess/QueryParser.java
+++ b/src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/preprocess/QueryParser.java
@@ -22,15 +22,25 @@ import org.opensearch.search.relevance.transformer.kendraintelligentranking.conf
 public class QueryParser {
   private static final Logger logger = LogManager.getLogger(QueryParser.class);
 
+  private static final String BODY_FIELD_REQUIRED_ERROR_MESSAGE =
+      "Property [" + BODY_FIELD + "] must be specified";
   private static final String FILED_MISMATCH_ERROR_MESSAGE =
       "Mismatch: Field configured in %s property [%s] is not present in query fields [%s]. Will not apply Kendra Intelligent Ranking.";
+  private static final String QUERY_PARSER_RESULT_LOG =
+      "Kendra Intelligent Ranker query parser fields for query type [%s]: bodyField: %s, titleField: %s";
+  private static final String QUERY_PARSER_RESULT_LOG_WITHOUT_TITLE =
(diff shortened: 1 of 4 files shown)


kevinawskendra (approved):

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/preprocess/QueryParser.java
```diff
@@ -22,15 +22,25 @@
...
   private static final Logger logger = LogManager.getLogger(QueryParser.class);
 
+  private static final String BODY_FIELD_REQUIRED_ERROR_MESSAGE =
+      "Property [" + BODY_FIELD + "] must be specified";
   private static final String FILED_MISMATCH_ERROR_MESSAGE =
       "Mismatch: Field configured in %s property [%s] is not present in query fields [%s]. Will not apply Kendra Intelligent Ranking.";
+  private static final String QUERY_PARSER_RESULT_LOG =
+      "Kendra Intelligent Ranker query parser result for query type [%s]: queryText: %s, bodyField: %s, titleField: %s";
+  private static final String QUERY_PARSER_RESULT_LOG_WITHOUT_TITLE =
+      "Kendra Intelligent Ranker query parser result for query type [%s]: queryText: %s, bodyField: %s";
```
msfroh: Do we want to enable logging for all queries sent to Kendra?

My big concern with that is a) the increase in the volume of stuff that we're logging and b) the risk of logging user queries without conscious action on the part of the admin.

mahitamahesh: Good point, especially with (b)

How about making this a debug log and removing the query text, but keeping body and title field names? It's difficult to debug which fields got sent to Kendra, especially since the title is optional.

msfroh: I think I'm okay with keeping it at INFO level, as long as we remove the query text.

The query text logging specifically is the part that I'm a little nervous about.

mahitamahesh: Addressed

msfroh (approved):

PR #24, opened 2022-11-10
Bug fix: Allow empty title field after tokenization

mahitamahesh:
### Description
If there are no tokens from tokenizing the title field, pass null to Kendra instead of empty list.
 
### Issues Resolved
N/A
 
### Check List
  - [x] All tests pass


src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/KendraIntelligentRanker.java
```diff
@@ -111,6 +111,10 @@ public SearchHits transform(final SearchHits hits,
         List<String> tokenizedTitle = null;
         if (queryParserResult.getTitleFieldName() != null) {
           tokenizedTitle = textTokenizer.tokenize(docSourceMap.get(queryParserResult.getTitleFieldName()).toString());
```
msfroh, line 113: I just checked and confirmed that the return value from `tokenize` is never null: https://github.com/opensearch-project/search-relevance/blob/main/src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/preprocess/TextTokenizer.java#L39

msfroh (approved):

PR #25, opened 2022-11-11
Add a Docker quickstart script for demo / test

msfroh:
Adding a Bash script to simplify setting up and launching local Docker containers with OpenSearch (with this plugin installed) and OpenSearch Dashboards, with the plugin configured to connect to Kendra ranking service.

### Description
Bash script does the following:
1. Create a `Dockerfile` that uses the OpenSearch 2.4.0 base image and installs this plugin.
2. If AWS credentials are found in the specified AWS profile, the current environment, or the default profile, then they get "pushed" into the OpenSearch keystore in the Docker image.
3. If requested, create a Kendra Intelligent Ranking rescore execution plan and wait for it to become active.
4. Build the Docker image from `Dockerfile` and tag it (as `opensearch-with-ranking-plugin`).
5. Create a `docker-compose.yml` that launches two nodes:
    1. OpenSearch based on the image we just built, with config settings pointing to a Kendra Ranking rescore execution plan.
    2. OpenSearch-Dashboards 2.4.0
6. Run `docker-compose up` to bring up both nodes.

NOTE: I have tested this on a couple of hosts (running Linux and Mac OS) with temporary and long-lived AWS credentials. I still need to test it on an EC2 instance using instance profile credentials.

 ### Issues Resolved
N/A
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/helpers/search_processing_kendra_quickstart.sh
+++ b/helpers/search_processing_kendra_quickstart.sh
@@ -1,0 +1,484 @@
+#!/bin/bash
+
+set -o errexit
+set -o errtrace
+set -o pipefail
+set -o nounset
+
+# Some useful constants
+readonly DOCKER_IMAGE_TAG="opensearch-with-ranking-plugin"
+readonly OPENSEARCH_VERSION="2.4.0"
+
+# Uncomment the following for prerelease testing:
+readonly SEARCH_PROCESSOR_PLUGIN_URL="https://github.com/msfroh/search-relevance/releases/download/v2.4.0-alpha/search-processor-disable-ssl-validation.zip"
+
+#
+# Set default values for OpenSearch (+Dashboards) image tags and plugin URL.
+# We can pass them from outside to override these settings for testing purposes.
+#
+if [ -z "${OPENSEARCH_IMAGE_TAG:-}" ]; then
+  OPENSEARCH_IMAGE_TAG="opensearchproject/opensearch:${OPENSEARCH_VERSION}"
+fi
+if [ -z "${OPENSEARCH_DASHBOARDS_IMAGE_TAG:-}" ]; then
+  OPENSEARCH_DASHBOARDS_IMAGE_TAG="opensearchproject/opensearch-dashboards:${OPENSEARCH_VERSION}"
+fi
+if [ -z "${SEARCH_PROCESSOR_PLUGIN_URL:-}" ]; then
+  SEARCH_PROCESSOR_PLUGIN_URL="https://github.com/opensearch-project/search-relevance/releases/download/${OPENSEARCH_VERSION}/search-processor.zip"
+fi
+
+function print_help() {
+  cat << EOF
+Usage: $0 [-p <execution_plan_id>] [-r <region>] [-e <kendra_ranking_endpoint>]
+        [--profile <AWS profile name>] [--create-execution-plan]
+  -p | --execution-plan-id          The ID returned from Kendra Intelligent Ranking service
+                                    from the call to CreateRescoreExecutionPlan. Required if
+                                    --create-execution-plan is not set.
+  -r | --region                     The AWS region for the Kendra Intelligent Ranking
+                                    service endpoint. If not specified, will read from the
+                                    AWS CLI for the default profile.
+  -e | --kendra-ranking-endpoint    The URL for the Kendra Intelligent Ranking service
+                                    endpoint. If not specified, a value based on the
+                                    region will be used.
+  --profile                         The AWS profile to use for credentials. If not set, then
+                                    the script will try first to use credentials from the
+                                    environment, then from the default AWS profile.
+  --create-execution-plan           If set, then the supplied AWS credentials will be used to
+                                    create a Kendra Intelligent Ranking execution plan with
+                                    1 capacity unit. NOTE: You will be charged for provisioned
+                                    execution plan.
+
+  NOTE: If the --profile option is not specified, the script will attempt to read AWS 
+  credentials (access/secret key, optional session token) from environment variables, 
+  and then from the default AWS profile, in order to pass them to the OpenSearch keystore
+  to be used to connect to the Kendra Intelligent ranking service.
+  If no credentials are found, the script WILL NOT pass credentials to the OpenSearch
+  keystore. When running a reranking request, the ranking plugin may rely on
+  instance profile credentials delivered through the EC2 metadata service, or credentials
+  from the ECS metadata service.
+EOF
+}
+
+#
+# Parse and validate arguments
+#
+
+while [ "$#" -gt 0 ]; do
+    case $1 in
+        -p | --execution-plan-id )
+            shift
+            EXECUTION_PLAN_ID=$1
+            shift
+            ;;
+        -r | --region )
+            shift
+            AWS_REGION=$1
+            shift
+            ;;
+        -e | --kendra-ranking-endpoint )
+            shift
+            KENDRA_RANKING_ENDPOINT=$1
+            shift
+	    ;;
+        -h | --help )
+	    print_help
+            exit 0
+            ;;
+	--profile )
+	    shift
+	    AWS_PROFILE=$1
+	    shift
+	    ;;
+        --create-execution-plan )
+            shift
+            CREATE_EXECUTION_PLAN=1
+            ;;
+    esac
+done
+
+FAILED_VALIDATION=0
+if [ -z "${EXECUTION_PLAN_ID:-}" ] && [ -z "${CREATE_EXECUTION_PLAN:-}" ]; then
+  >&2 echo "Missing argument [-p | --execution-plan] or --create-execution-plan"
+  FAILED_VALIDATION=1
+elif [ -n "${EXECUTION_PLAN_ID:-}" ] && [ -n "${CREATE_EXECUTION_PLAN:-}" ]; then
+  >&2 echo "You cannot specify both [-p | --execution-plan] and --create-execution-plan"
+  FAILED_VALIDATION=1
+fi
+
+# End of required arguments. If any are missing, print help and exit.
+if [ "${FAILED_VALIDATION}" == "1" ]; then
+  echo
+  print_help
+  exit 1
+fi
+
+#
+# Determine which credentials and region to use. By the end of this block, all specified 
+# credentials will be loaded into environment variables (or we fail with an explanatory 
+# error message).
+#
+if [ -n "${AWS_PROFILE:-}" ]; then
+  # Load everything from the specified profile
+  AWS_ACCESS_KEY_ID=$(aws --profile ${AWS_PROFILE} configure get aws_access_key_id || echo)
+  AWS_SECRET_ACCESS_KEY=$(aws --profile ${AWS_PROFILE} configure get aws_secret_access_key || echo)
+  AWS_SESSION_TOKEN=$(aws --profile ${AWS_PROFILE} configure get aws_session_token || echo)
+  if [ -z "${AWS_ACCESS_KEY_ID:-}" ] || [ -z "${AWS_SECRET_ACCESS_KEY:-}" ]; then
+    >&2 echo "Unable to load credentials from profile ${AWS_PROFILE}"
+    exit 1
+  elif [ -z "${AWS_SESSION_TOKEN}" ]; then
+    echo "Using AWS credentials (aws_access_key_id and aws_secret_access_key) from profile ${AWS_PROFILE}"
+  else 
+    echo "Using AWS credentials (aws_access_key_id, aws_secret_access_key, and aws_session_token) from profile ${AWS_PROFILE}"
+  fi
+  if [ -z "${AWS_REGION:-}" ]; then
+    AWS_REGION=$(aws --profile ${AWS_PROFILE} configure get region || echo)
+    if [ -n "${AWS_REGION:-}" ]; then
+      echo "Using AWS region ${AWS_REGION} from profile ${AWS_PROFILE}"
+    else
+      >&2 echo "Argument [-r | --region] not specified and unable to infer region from profile ${AWS_PROFILE}"
+      exit 1
+    fi
+  fi
+else
+  # No profile set
+  if [ -z "${AWS_ACCESS_KEY_ID:-}" ]; then
+    if [ -z "${AWS_SECRET_ACCESS_KEY:-}" ]; then
+      echo "No profile set and no credentials in environment. Trying to load default profile credentials."
+      AWS_ACCESS_KEY_ID=$(aws configure get aws_access_key_id || echo)
+      AWS_SECRET_ACCESS_KEY=$(aws configure get aws_secret_access_key || echo)
+      AWS_SESSION_TOKEN=$(aws configure get aws_session_token || echo)
+      if [ -z "${AWS_ACCESS_KEY_ID:-}" ] || [-z "${AWS_SECRET_ACCESS_KEY}" ]; then
+        echo "Unable to load credentials from default profile. No credentials will be passed to the OpenSearch keystore."
+        echo "OpenSearch will use the default credential provider chain to access Kendra, which may rely on EC2 instance"
+        echo "profile credentials or credentials from ECS metadata service."
+      elif [ -z "${AWS_SESSION_TOKEN}" ]; then
+        echo "Using AWS credentials (aws_access_key_id and aws_secret_access_key) from default profile."
+      else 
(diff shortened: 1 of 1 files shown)


mahitamahesh (approved):

helpers/search_processing_kendra_quickstart.sh
```diff
@@ -0,0 +1,174 @@
...
+set -o pipefail
+set -o nounset
+
+# Some useful constants
+readonly DOCKER_IMAGE_TAG="opensearch-with-ranking-plugin"
+readonly OPENSEARCH_VERSION="2.4.0"
+
+function print_help() {
+  echo "Usage: $0 -p <execution_plan_id> [-r <region>] [-e <kendra_ranking_endpoint>]"
+  echo "  -p | --execution-plan-id          The ID returned from Kendra Intelligent Ranking service"
```
rstrahan: Can we / should we make this optional, and have the script create an execution plan automatically (with defaults) if one isn't provided? Since there's a cost implication, require this to be explcitly requested, e.g. require an explicit arg like "--create-kendra-ranking-rescore-execution-plan" to have script create the execution plan automatically.

helpers/search_processing_kendra_quickstart.sh
```diff
@@ -0,0 +1,174 @@
...
+  echo "                                    from the call to CreateRescoreExecutionPlan"
+  echo "  -r | --region                     The AWS region for the Kendra Intelligent Ranking"
+  echo "                                    service endpoint. If not specified, will read from the"
+  echo "                                    AWS CLI for the default profile."
+  echo "  -e | --kendra-ranking-endpoint    The URL for the Kendra Intelligent Ranking service"
+  echo "                                    endpoint. If not specified, a value based on the"
+  echo "                                    region will be used."
+  echo
+  echo "  NOTE: The script will attempt to read AWS credentials (access/secret key)"
+  echo "  from environment variables, and then from the default AWS profile, in order to"
```
rstrahan: Can you add support for `--profile` rather than assume use of default profile only?

helpers/search_processing_kendra_quickstart.sh
```diff
@@ -0,0 +1,174 @@
...
+    image: opensearchproject/opensearch-dashboards:${OPENSEARCH_VERSION}
+    container_name: opensearch-dashboards
+    ports:
+      - 5601:5601
+    environment:
+      OPENSEARCH_HOSTS: '["https://opensearch-node:9200"]'
+    networks:
+      - opensearch-net
+EOF
+docker-compose up
```
rstrahan: Suggest we output additional helpful messages at the end of this script, e.g.
- opensearch endpoint, and simple curl command for testing it
- URL for dashboards
- command to use to stop and restart the docker node
- how to access the opensearch server log messages

helpers/search_processing_kendra_quickstart.sh
msfroh: That's a good idea!  I'll add that in my next revision.

helpers/search_processing_kendra_quickstart.sh
msfroh: Will do!

helpers/search_processing_kendra_quickstart.sh
msfroh: For now, the docker-compose command is followed by a flurry of output for both containers (as they both flush to stdout on the current shell).  That's actually been kind of handy for me for debugging.

Unfortunately, it means that any helpful messages would get swept away.

I could switch this to `docker-compose up -d` to run in the background and offer those getting-started messages. Alternatively, instead of actually starting the containers, I could suggest that they go into the directory and run `docker-compose up` themselves, and suggest what they can do once the nodes are started.

I think the `-d` option makes more sense. I'll look into adding that in my next revision.

helpers/search_processing_kendra_quickstart.sh
msfroh: Will do!

helpers/search_processing_kendra_quickstart.sh
rstrahan: I agree.. Start the containers with -d, and exit the script with helpful messages. Thanks!

helpers/search_processing_kendra_quickstart.sh
rstrahan: Another thought.. Since creating the execution plan will require the very latest version of the CLI (with support for kendra-ranking) can the script defend against an old version too? ie check the version, and if kendra-ranking isn't supported, then output link to cli update instructions?  That'll cover one more thing that could go wrong :)

helpers/search_processing_kendra_quickstart.sh
```diff
@@ -0,0 +1,228 @@
...
+    echo "Could not infer aws_access_key_id from environment or profile. Not adding to OpenSearch keystore."
+  fi
+fi
+if [ -z "${AWS_SECRET_ACCESS_KEY:-}" ]; then
+  AWS_SECRET_ACCESS_KEY=$(aws configure get aws_secret_access_key || echo "")
+  if [ -z "${AWS_SECRET_ACCESS_KEY:-}" ]; then
+    echo "Could not infer aws_secret_access_key from environment or profile. Not adding to OpenSearch keystore."
+  fi
+fi
+if [ -z "${AWS_SESSION_TOKEN:-}" ]; then
```
rstrahan: Michael - this logic is causing problems when user has configured environment variables for AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY on a system where the aws credentials file contains temporary credentials ( aws_session_token).  The script correctly reads the  AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY from the environment, BUT the code here will also set AWS_SESSION_TOKEN to the value stored in the credentials file. This session token is invalid, and results in an OpenSearch "security token is invalid" exception.  
When using environment variables for AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY the script should use only an environment variable for AWS_SESSION_TOKEN, or none.

helpers/search_processing_kendra_quickstart.sh
```diff
@@ -0,0 +1,228 @@
...
+
+# End of required arguments. If any are missing, print help and exit.
+if [ "${ERROR}" == "1" ]; then
+  echo
+  print_help
+  exit 1
+fi
+
+if [ -z "${KENDRA_RANKING_ENDPOINT:-}" ]; then
+  KENDRA_RANKING_ENDPOINT="https://kendra-ranking.${AWS_REGION}.amazonaws.com"
```
kevinawskendra: We are actually changing the default endpoint to follow the format `https://kendra-ranking.${AWS_REGION}.api.aws`. Can you please update this? Thanks!

helpers/search_processing_kendra_quickstart.sh
msfroh: Good point!

I wasted a couple of hours on Monday wondering why the script wasn't working for me. It turned out that I had set the AWS_ACCESS_KEY_ID variable in the shell in that particular terminal window and it was getting mixed with my temporary (secret key + session token) credentials.

I'll fix the logic to either:
1. use any and all available values from environment variables, 
2. use any and all available values from the (explicitly specified or implicitly default) profile, or 
3. add no credentials to the keystore (so at runtime, fall back to the DefaultAWSCredentialsProviderChain, which is great if you're using EC2 instance profile or ECS metadata credentials).

helpers/search_processing_kendra_quickstart.sh
msfroh: Thanks, fixed!

helpers/search_processing_kendra_quickstart.sh
msfroh: I added some (hopefully) helpful text at the end.

It got a little long, so it outputs the text to a README file and then outputs that, with the last instruction being "If you ever want to read this text again, I've written it `<working_directory>/README`".

I covered your 4 suggestions, pointed to the docs for indexing, and explained how to add Kendra ranking to a query. I'm open to other ideas for things that might be useful.

helpers/search_processing_kendra_quickstart.sh
msfroh: I added support for creating the execution plan.

It checks for kendra-ranking support based on the exit code for `aws kendra-ranking help` (which works -- I tried with and without the kendra-ranking service model added to my CLI).

To avoid bringing up the OpenSearch containers before the execution plan is ready, it sits in a loop describing the execution plan every 30 seconds until the status becomes "ACTIVE" before proceeding.

rstrahan:
Brilliant!! Will test it tomorrow (later today!)
Thanks!!

Sent from Workspace ONE Boxer

On Nov 17, 2022 2:02 AM, msfroh ***@***.***> wrote:

@msfroh commented on this pull request.

________________________________

In helpers/search_processing_kendra_quickstart.sh<https://github.com/opensearch-project/search-relevance/pull/25#discussion_r1024817787>:

> +    ports:
+      - 9200:9200
+      - 9600:9600
+    networks:
+      - opensearch-net
+  opensearch-dashboard:
+    image: opensearchproject/opensearch-dashboards:${OPENSEARCH_VERSION}
+    container_name: opensearch-dashboards
+    ports:
+      - 5601:5601
+    environment:
+      OPENSEARCH_HOSTS: '["https://opensearch-node:9200"]'
+    networks:
+      - opensearch-net
+EOF
+docker-compose up


I added some (hopefully) helpful text at the end.

It got a little long, so it outputs the text to a README file and then outputs that, with the last instruction being "If you ever want to read this text again, I've written it <working_directory>/README".

I covered your 4 suggestions, pointed to the docs for indexing, and explained how to add Kendra ranking to a query. I'm open to other ideas for things that might be useful.

helpers/search_processing_kendra_quickstart.sh
```diff
@@ -0,0 +1,419 @@
...
+    export AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY
+    if [ -n "${AWS_SESSION_TOKEN:-}" ]; then
+      export AWS_SESSION_TOKEN
+    fi
+  fi
+  EXECUTION_PLAN_ID=$(aws ${AWS_CLIENT_OPTS:-} --region ${AWS_REGION} --endpoint-url ${KENDRA_RANKING_ENDPOINT} \
+    kendra-ranking create-rescore-execution-plan --name "TestPlan" \
+      --description "Plan created by search_processing_kendra_quickstart.sh" \
+    --capacity-units RescoreCapacityUnits=1 --output text --query Id)
+  echo "Created execution plan with ID ${EXECUTION_PLAN_ID}. "
```
rstrahan: The CLI command above has problems @msfroh 
1.   `--capacity-units RescoreCapacityUnits=1` results in error 
```
Parameter validation failed:
Missing required parameter in CapacityUnits: "QueryCapacityUnits"
Unknown parameter in CapacityUnits: "RescoreCapacityUnits", must be one of: QueryCapacityUnits
```
It's unnecessary.. please just remove it, since the default is '1' anyway.

2. When using a custom endpoint, can you also add `--no-verify-ssl` to prevent SSL failure of the command line.. both here, and in the decsribe-rescore-execution-plan loop below

3. Hardcoded name `--name "TestPlan"` causes problem when script is run a second time, and execution plan exists.  
```
An error occurred (ConflictException) when calling the CreateRescoreExecutionPlan operation: The same request is currently associated with a different client token. To create another execution plan, please change a request parameter and try your request again.
```
To address this please modify logic to catch this error, and proceed to reuse the existing execution plan (you'll need to retrieve its ID using list-rescore-execution-plans I think)

helpers/search_processing_kendra_quickstart.sh
```diff
@@ -0,0 +1,419 @@
...
+analyzed for rescoring and <title_field> is the name of an (optional) field 
+representing the title of your documents.
+
+Interact with the Docker containers using docker-compose from directory
+  $(pwd)
+
+Some helpful docker-compose commands:
+
+  docker-compose logs opensearch-node 
+                Outputs latest logs from the OpenSearch server
```
rstrahan, line 405: When I run this on my linux machine after completing the script I get:
```
$ docker-compose logs opensearch-node
no configuration file provided: not found
```

helpers/search_processing_kendra_quickstart.sh
```diff
@@ -0,0 +1,419 @@
...
+
+# 
+# Bring up the images
+#
+docker-compose up -d
+
+#
+# Output helpful "getting started" text.
+#
+cat >README <<EOF
```
rstrahan, line 442: Also include instructions for Cleanup
- remove docker containers
- delete execution plan

REQUEST - Actually - it will be much better to add an option to the script to completely cleanup all resources automatically. Say "--uninstall". Add confirmation (Yes/No) to avoid accidental deletion. Should delete
- execution plan
- docker containers
- the directory and files created during install

helpers/search_processing_kendra_quickstart.sh
msfroh: >Parameter validation failed:
Missing required parameter in CapacityUnits: "QueryCapacityUnits"
Unknown parameter in CapacityUnits: "RescoreCapacityUnits", must be one of: QueryCapacityUnits

The QueryCapacityUnits parameter was renamed to RescoreCapacityUnits a few days ago. You need the latest service model to get that change. I'll remove the argument if it defaults to 1, though.

> When using a custom endpoint, can you also add --no-verify-ssl to prevent SSL failure of the command line.. both here, and in the decsribe-rescore-execution-plan loop below

That's what the `${AWS_CLIENT_OPTS:-}` variable is for -- I'm running the command with `AWS_CLIENT_OPTS=--no-verify-ssl ./search_processing_kendra_quickstart.sh [...]`. As far as I know that should never be needed by real users, though. This is only because we're using pre-release endpoints with invalid certificates.

> Hardcoded name --name "TestPlan" causes problem when script is run a second time, and execution plan exists.

I think I can check if a plan with that name exists before trying to create it. If it's in a CREATING state, it can wait. If it's in ACTIVE state we can use it. If it has any other state, I think I would error out.

helpers/search_processing_kendra_quickstart.sh
msfroh: The script now generates its own cleanup script that shuts down Docker containers and deletes any created rescore execution plan.

The output looks like:

```
This script will shut down and clean up running Docker containers for OpenSearch and OpenSearch Dashboards.
And it will delete the Kendra Intelligent Ranking rescore execution plan with ID d08cb653-4a51-4f38-b765-75cd767b21f8
Are you sure? (yes/no)
yes
Stopping opensearch-node       ... done
Stopping opensearch-dashboards ... done
Removing opensearch-node       ... done
Removing opensearch-dashboards ... done
Removing network opensearch-kendra-ranking-dockertino_opensearch-net
Deleting rescore execution plan with ID d08cb653-4a51-4f38-b765-75cd767b21f8...
```

It will only delete stuff with an explicit `yes` answer:

```
This script will shut down and clean up running Docker containers for OpenSearch and OpenSearch Dashboards.
And it will delete the Kendra Intelligent Ranking rescore execution plan with ID d08cb653-4a51-4f38-b765-75cd767b21f8
Are you sure? (yes/no)
y
Are you sure? (yes/no)
maybe
Are you sure? (yes/no)
sometimes
Are you sure? (yes/no)
no
```

helpers/search_processing_kendra_quickstart.sh
msfroh: 1. Removed the `--capacity-units` param.
2. Use `AWS_CLIENT_OPTS=--no-verify-ssl`. (Not documenting this because AFAIK it should never be an issue for real AWS users and we don't want real AWS users disabling SSL verification.)
3. Added check for existing rescore execution plans named "TestPlan". (You can have multiple plans with the same name as long as some other parameter has changed, like the description or capacity units.)
    1. It will reuse the first plan with status "CREATING" or "ACTIVE".
    2. It ignores any plan with the name "DELETING" (since that doesn't prevent creating a new plan with those parameters).
    3. It fails if it finds a plan named "TestPlan" and some other status (since I don't know what that means).

Between this and the cleanup script, I'm finding this whole thing much friendlier. Let me know how it works for you.

msfroh:
Oops -- I marked this as "ready for review", but I want to hold off until we have a released plugin zip. The default plugin zip location is still a 404.

helpers/search_processing_kendra_quickstart.sh
rstrahan: All sounds great.. I'll give it another whirl today. Thanks!

macohen (approved):
We should plan to deprecate this after the 2.5 release is out and replace it with instructions for plugin installation.

PR #30, opened 2022-11-22
Add test coverage for SearchActionFilter

### LINKED ISSUE #31: Improve test coverage
This project currently (November 22, 2022) has around 21% line coverage from tests. 

We should increase that significantly to get more confidence on future changes (and maybe uncover some lurking bugs along the way).

I'm going to suggest a target of 80% before we can call this issue "done".

msfroh:

### Description
Adds unit tests for SearchActionFilter class (plus OpenSearchClient as collateral impact).

### Issues Resolved
#31 won't actually be "resolved", but this contributes.
 
### Check List
  - [X] All tests pass

### DIFF
--- a/src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
+++ b/src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
@@ -35,6 +35,7 @@ import org.opensearch.common.io.stream.BytesStreamOutput;
 import org.opensearch.common.io.stream.NamedWriteableAwareStreamInput;
 import org.opensearch.common.io.stream.NamedWriteableRegistry;
 import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.settings.Setting;
 import org.opensearch.search.SearchHit;
 import org.opensearch.search.SearchHits;
 import org.opensearch.search.aggregations.InternalAggregations;
@@ -107,7 +108,8 @@ public class SearchActionFilter implements ActionFilter {
       // Source is returned in response hits by default. If disabled by the user, overwrite and enable
       // in order to access document contents for reranking, then suppress at response time.
       boolean suppressSourceOnResponse = false;
-      if (searchRequest.source().fetchSource() != null && !searchRequest.source().fetchSource().fetchSource()) {
+      if (searchRequest.source() != null && searchRequest.source().fetchSource() != null &&
+              !searchRequest.source().fetchSource().fetchSource()) {
         searchRequest.source().fetchSource(true);
         suppressSourceOnResponse = true;
       }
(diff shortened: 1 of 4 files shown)


src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
```diff
@@ -107,7 +108,8 @@ public <Request extends ActionRequest, Response extends ActionResponse> void app
       // Source is returned in response hits by default. If disabled by the user, overwrite and enable
       // in order to access document contents for reranking, then suppress at response time.
       boolean suppressSourceOnResponse = false;
-      if (searchRequest.source().fetchSource() != null && !searchRequest.source().fetchSource().fetchSource()) {
+      if (searchRequest.source() != null && searchRequest.source().fetchSource() != null &&
+              !searchRequest.source().fetchSource().fetchSource()) {
```
macohen, line 112: Is it a problem that we have to reach so deep into the SearchRequest?  Not saying that we can do anything different here...

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
```diff
@@ -143,14 +145,13 @@ private List<ResultTransformerConfiguration> getResultTransformerConfigurations(
...
             .stream()
-            .map(s -> s.getKey())
-            .collect(Collectors.toList()))
-        .flatMap(Collection::stream)
-        .collect(Collectors.toList())
-        .toArray(new String[0]);
+            .map(t -> t.getTransformerSettings()
+                    .stream()
+                    .map(Setting::getKey)
+                    .collect(Collectors.toList()))
```
macohen, line 152: Why this change? Is it because collect was being called twice unnecessarily?

src/main/java/org/opensearch/search/relevance/configuration/SearchConfigurationExtBuilder.java
```diff
@@ -124,8 +124,10 @@ public int hashCode() {
     return Objects.hash(this.getClass(), this.resultTransformerConfigurations);
   }
 
-  public void setResultTransformers(final List<ResultTransformerConfiguration> resultTransformerConfigurations) {
-    this.resultTransformerConfigurations = resultTransformerConfigurations; }
+  public SearchConfigurationExtBuilder setResultTransformers(final List<ResultTransformerConfiguration> resultTransformerConfigurations) {
```
macohen, line 127: is there a different standard now for creating a "set" method that returns something? Should this be named differently?

macohen (commented):
Just had a few questions, but looks good overall.

src/main/java/org/opensearch/search/relevance/configuration/SearchConfigurationExtBuilder.java
msfroh: I normally write builders with `withXXX` methods for chaining, but the longstanding convention with builders in the codebase uses `setXXX` that returns the builder.

I disagree with the convention, but I need to follow it. (Well, I guess I don't *need* to, but I think consistency is more important in this case.)

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
msfroh: Yeah... IntelliJ flagged it as a warning, effectively saying "Why are you taking two steps to convert this stream to an array when you can just do it in one step?" With a quick Alt-Enter, it fixed it for me.

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
msfroh: Yeah... this is just how the core SearchRequest class is structured:

* SearchRequest holds (mostly optional) high-level flow-control type stuff (routing param, scroll state, timeout), plus a reference to the SearchSourceBuilder (which never actually builds a SearchSource, but 🤷 ).
* SearchSourceBuilder holds most of what we think of as "the search request" -- the query, highlight options, sort options, from+size, etc. including the FetchSourceContext.
* The FetchSourceContext effectively represents return field options -- `fetchSource` is a boolean where `true` means "Yes, return all fields" and `false` means "No, return no fields". You can also specify includes/excludes (which may include wildcards) to fetch a subset of fields.

macohen (approved):

PR #32, opened 2022-11-23
Add pagination / document limit logic for reranking

msfroh:
### Description
This change augments https://github.com/opensearch-project/search-relevance/pull/26 by adding unit tests.

* Add `preprocessRequest` interface to ResultTransformer.
* Override `from` and `size` parameters during preprocess, to fetch more than the default 10 hits from OpenSearch (improved accuracy if the ranker can access more documents)
* Limit the number of documents sent to Kendra Ranking (Will make this number configurable in a follow up)
* Populate response documents from re-ranked hits (upto Kendra limit) and OpenSearch hits beyond the limit

### Issues Resolved
N/A
 
### Check List
  - [X] All tests pass


src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/KendraIntelligentRanker.java
```diff
@@ -126,6 +127,11 @@ public SearchHits transform(final SearchHits hits,
         kendraConfig.getProperties().getBodyFields(),
         kendraConfig.getProperties().getTitleFields());
     if (queryParserResult == null) {
+      // Unknown query type or query does not reference body field
+      return hits;
+    }
+    if (hits.getHits().length == 0) {
```
mahitamahesh: Nit: If we check this right up front, we can avoid the query parser step too

src/test/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/KendraIntelligentRankerTests.java
```diff
@@ -0,0 +1,210 @@
...
+    public void testShouldNotTransformWithoutSource() {
+        KendraIntelligentRanker ranker = new KendraIntelligentRanker(buildMockHttpClient());
+        SearchRequest originalRequest = new SearchRequest();
+        boolean shouldTransform = ranker.shouldTransform(originalRequest, new KendraIntelligentRankingConfiguration());
+        assertFalse(shouldTransform);
+    }
+
+    public void testShouldNotTransformWithoutQuery() {
+        KendraIntelligentRanker ranker = new KendraIntelligentRanker(buildMockHttpClient());
+        SearchRequest originalRequest = new SearchRequest();
```
mahitamahesh: We should set the SearchSourceBuilder with query(null) here

msfroh: Oops! Thanks

I think I copy/pasted the test above, got distracted, then came back and started writing the next test. 😁

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/KendraIntelligentRanker.java
msfroh: Good catch, thanks!

mahitamahesh (approved):

msfroh:
Yay for randomized testing! 

My earlier commit failed on the case where `testTransformHits` randomly selected zero as a docLimit and zero additional hits. That's already covered by the earlier `testTransformHits` test case so we can safely skip it. 

Still, it's nice to know that we're considering many possibilities.

macohen (approved):

PR #33, opened 2022-11-23
Modifying logic to generate passages from search result document

mahitamahesh:
Signed-off-by: Mahita Mahesh <mahitam@amazon.com>

### Description
Compared to the previous logic, this change adds:
- More filters and transformations during tokenization
- Simplifies sentence splitting logic
- Generates passages with overlap
 
### Issues Resolved
N/A
 
### Check List
  - [x] All tests pass


src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/preprocess/TextTokenizer.java
```diff
@@ -26,26 +30,83 @@ public class TextTokenizer {
...
+      return false;
+    }
+    return true;
+  }
+
+  String removeInWordPunctuation(String token) {
+    if (token == null) {
+      return null;
+    }
+    return token.replaceAll(PUNCTUATIONS_REGEX_PATTERN, "");
```
msfroh: Can you compile the regex once into a `static final Pattern`, and replace this with:

```
PUNCTUATIONS_REGEX.matcher(token).replaceAll("");
```

As written, this will compile the regex on every token.

mahitamahesh: Ah! I changed the one above and missed this one.. thanks for catching!

msfroh (approved):

Issue #36, opened 2022-11-23
[RFC] Additional Reranker Implementations Requested


macohen:
We want to be sure we call ourselves out as per our [principles for development](https://opensearch.org/about.html#principles-for-development). Specifically, because we are releasing this plugin with AWS Kendra as the first integrated reranker, we are unfortunately violating the "A level playing field" principle. We do see the Kendra re-ranking integration as simply the first re-ranker integration and want to add more as soon as possible. In calling this out, we are also asking the community for comments here, PRs on the plugin for additional integrations if interested, as well as cooperative work to add an additional re-ranker so we can refactor the code and generalize it.

macohen:
The background for this feature is available in #4.

jmazanec15:
Hi @macohen one use case I can think of would be the ability to re-rank post query phase inside the cluster. For instance, a re-ranking plugin inside the cluster that uses ML Commons to host and execute model could be beneficial. 

Would it be possible to make reranking final results after a query phase a plugin? So that implementers would not have to implement the action filter directly?

macohen:
@jmazanec15 that's the intent of the search-processor so this should fit. How do users do this with ML Commons today? Can you provide some more information so we can verify. We also do have an RFC for the overall architecture of search-processor coming in January so there should be much more to discuss then.

jmazanec15:
@macohen Currently, I am not aware of any users using ml-commons to do re-ranking today. However, I could see it being extended to this use case. ML Commons provides an extendible interface for model inference workloads. (here is the [client](https://github.com/opensearch-project/ml-commons/blob/main/client/src/main/java/org/opensearch/ml/client/MachineLearningClient.java#L47)). I imagine a re-ranker plugin could be implemented that would use a reranking model hosted with ml-commons to re-rank the top K search results.

macohen:
Closing. Please comment on #80 if you have ideas for additional reranker implementations.

PR #38, opened 2022-12-02
Bug fixes and refactoring

msfroh:
1. Shallow clone SearchResponse in SearchActionFilter. The only properties that we're currently changing are SearchHit and timeTookMillis. This is also less brittle as we don't try to implement deserialization ourselves.
2. KendraIntelligentRanker doesn't transform if the required client settings (endpoint + execution plan ID) are missing.
3. A lot of refactoring to prepare for other transformers. Where we previously had logic of the form `if kendra_intelligent_ranking then do Kendra intelligent ranking`, now the APIs support arbitrary named transformers. This required updates to tests and configuration code.
4. Added tests and fixes for SearchConfigurationExtBuilder.

### Description
See commit message above.
 
### Issues Resolved
* Refactoring is a prerequisite for https://github.com/opensearch-project/search-processor/issues/36
* Additional tests contribute to https://github.com/opensearch-project/search-processor/issues/31
 
### Check List
  - [x] All tests pass


src/main/java/org/opensearch/search/relevance/configuration/ConfigurationUtils.java
```diff
@@ -27,16 +27,16 @@ public class ConfigurationUtils {
...
+                                                                                                 Map<String, ResultTransformer> resultTransformerMap) {
     List<ResultTransformerConfiguration> indexLevelConfigs = new ArrayList<>();
 
     if (settings != null) {
       if (settings.getGroups(RESULT_TRANSFORMER_SETTING_PREFIX) != null) {
-        for (Map.Entry<String, Settings> resultTransformer : settings.getGroups(RESULT_TRANSFORMER_SETTING_PREFIX).entrySet()) {
-          ResultTransformerType resultTransformerType = ResultTransformerType.fromString(resultTransformer.getKey());
-          if (ResultTransformerType.KENDRA_INTELLIGENT_RANKING.equals(resultTransformerType)) {
-            indexLevelConfigs.add(new KendraIntelligentRankingConfiguration(resultTransformer.getValue()));
+        for (Map.Entry<String, Settings> tranformerSettings : settings.getGroups(RESULT_TRANSFORMER_SETTING_PREFIX).entrySet()) {
```
msfroh: Typo: `tranformerSettings` -> `transformerSettings`

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
```diff
@@ -43,27 +31,43 @@
...
+
 public class SearchActionFilter implements ActionFilter {
     private static final Logger logger = LogManager.getLogger(SearchActionFilter.class);
 
     private final int order;
 
     private final NamedWriteableRegistry namedWriteableRegistry;
-    private final Map<ResultTransformerType, ResultTransformer> supportedResultTransformers;
+    private final Map<String, ResultTransformer> resultTransformerMap = new HashMap<>();
+    private final Map<String, ResultTransformerConfigurationFactory> configurationFactoryMap = new HashMap<>();
```
msfroh: This is unused. 

I was thinking of constructing configs separate from selecting the transformers, but then I never followed through. (Instead I just added a `getConfigurationFactory()` method to the `ResultTransfomer` interface.)

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
```diff
@@ -154,7 +158,7 @@ private List<ResultTransformerConfiguration> getResultTransformerConfigurations(
         }
 
         // Fetch all index settings for this plugin
-        String[] settingNames = supportedResultTransformers.values()
+        String[] settingNames = resultTransformerMap.values()
                 .stream()
                 .map(t -> t.getTransformerSettings()
                         .stream()
```
msfroh: I just realized that this can be simplified down to one `flatMap` like:

```
        String[] settingNames = resultTransformerMap.values()
                .stream()
                .flatMap(t -> t.getTransformerSettings()
                        .stream()
                        .map(Setting::getKey))
                .toArray(String[]::new);
```

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
```diff
@@ -230,52 +239,22 @@ public void onResponse(final Response response) {
...
+                    final long tookInMillis = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startTime);
+                    final SearchResponse newResponse = new SearchResponse(internalResponse, searchResponse.getScrollId(),
+                            searchResponse.getTotalShards(), searchResponse.getSuccessfulShards(),
+                            searchResponse.getSkippedShards(), tookInMillis, searchResponse.getShardFailures(),
+                            searchResponse.getClusters());
                     listener.onResponse((Response) newResponse);
-
-                    // TODO: Change this to a metric
-                    logger.info("Result transformer operations overhead time: {}ms",
-                            tookInMillis - searchResponse.getTook().getMillis());
```
msfroh, line 278: Is there value in logging the total time spent on transforms? 

We output the time for individual transformers above. That doesn't cover the time spent e.g. fetching index settings.

src/main/java/org/opensearch/search/relevance/configuration/ResultTransformerConfigurationFactory.java
```diff
@@ -0,0 +1,41 @@
...
+     */
+    ResultTransformerConfiguration configureFromIndexSettings(Settings indexSettings);
+
+    /**
+     * Build configuration from search request.
+     * @param parser a search request parser pointing to a child node under result transformers scoped to this result
+     *               transformer's name.
+     * @return a transformer configuration based on the parameters specified in the search request.
+     */
+    ResultTransformerConfiguration configureFromSearchRequest(XContentParser parser) throws IOException;
```
msfroh: TODO: Rename this. 

When I first wrote this method I was thinking "Oh, this is for the search request use-case", but it could apply to any XContent serialization.

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
macohen: I think if we can log so we're separating time spent in a transformer vs time spent fetching index settings and doing other internal opensearch operations, that would be good because then a transformer that goes haywire could be investigated and tracked better. If you still have doubts, can you describe more?

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/client/KendraHttpClient.java
```diff
@@ -34,6 +34,7 @@
 import java.security.AccessController;
```
noCharger, line 34:
```
The type AccessController has been deprecated since version 17 and marked for removalJava(16778626)
```
We should probably revisit this in the near future.

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/client/KendraHttpClient.java
```diff
@@ -50,45 +51,53 @@ public class KendraHttpClient {
...
-    amazonHttpClient = AccessController.doPrivileged((PrivilegedAction<AmazonHttpClient>) () -> new AmazonHttpClient(new ClientConfiguration()));
-    errorHandler = new SimpleAwsErrorHandler();
-    responseHandler = new SimpleResponseHandler();
-    aws4Signer = new AWS4Signer();
-    aws4Signer.setServiceName(KENDRA_RANKING_SERVICE_NAME);
-    aws4Signer.setRegionName(clientSettings.getServiceRegion());
-    objectMapper = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
     serviceEndpoint = clientSettings.getServiceEndpoint();
     executionPlanId = clientSettings.getExecutionPlanId();
+    if (StringUtils.isNotEmpty(serviceEndpoint) && StringUtils.isNotEmpty(executionPlanId)) {
```
noCharger: Good use of `StringUtils.isNotEmpty()`. We can reuse `isValid()` implemented below.


Shall we add more unit test cases to this condition? The only case I saw is:

```
        KendraHttpClient invalidClient = new KendraHttpClient(KendraClientSettings.getClientSettings(emptySettings));
```

Also, shall we take negation of it to reduce nest layers of `if`?

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
```diff
@@ -43,27 +31,39 @@
 import org.opensearch.search.fetch.subphase.FetchSourceContext;
 import org.opensearch.search.internal.InternalSearchResponse;
 import org.opensearch.search.profile.SearchProfileShardResults;
-import org.opensearch.search.relevance.configuration.ConfigurationUtils;
 import org.opensearch.search.relevance.client.OpenSearchClient;
+import org.opensearch.search.relevance.configuration.ConfigurationUtils;
 import org.opensearch.search.relevance.configuration.ResultTransformerConfiguration;
+import org.opensearch.search.relevance.configuration.ResultTransformerConfigurationFactory;
```
noCharger: nit: Unused import

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
```diff
@@ -43,27 +31,39 @@
...
 import org.opensearch.search.relevance.transformer.ResultTransformer;
-import org.opensearch.search.relevance.transformer.ResultTransformerType;
-import org.opensearch.search.suggest.Suggest;
 import org.opensearch.tasks.Task;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
```
noCharger: nit: Unused import

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/KendraIntelligentRanker.java
```diff
@@ -42,19 +32,31 @@
...
+    private static final int MIN_PASSAGE_LENGTH_IN_TOKENS = 100;
+    private static final int MAX_PASSAGE_COUNT = 10;
+    private static final int TITLE_TOKENS_TRIMMED = 15;
+    private static final int BODY_PASSAGE_TRIMMED = 200;
+    private static final double BM25_B_VALUE = 0.75;
+    private static final double BM25_K1_VALUE = 1.6;
+    private static final int TOP_K_PASSAGES = 3;
 
     private static final Logger logger = LogManager.getLogger(KendraIntelligentRanker.class);
+    public static final String NAME = "kendra_intelligent_ranking";
```
noCharger: I would put it right after the function signature since it's more organized and easier to read.

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/KendraIntelligentRanker.java
```diff
@@ -194,16 +207,16 @@ public SearchHits transform(final SearchHits hits,
...
+                maxScore = Math.max(maxScore, rescoreResultItem.getScore());
+                newSearchHits.add(searchHit);
             }
             // Add remaining hits to response, which are already sorted by OpenSearch score
             for (int i = numberOfHitsToRerank; i < originalHits.size(); ++i) {
-              newSearchHits.add(originalHits.get(i));
+                newSearchHits.add(originalHits.get(i));
             }
-            return new SearchHits(newSearchHits.toArray(new SearchHit[newSearchHits.size()]), hits.getTotalHits(), maxScore);
+            return new SearchHits(newSearchHits.toArray(new SearchHit[0]), hits.getTotalHits(), maxScore);
```
noCharger, line 219: I would like to know more about `newSearchHits.size()` -> `0`

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/configuration/KendraIntelligentRankingConfigurationFactory.java
```diff
@@ -0,0 +1,46 @@
...
+import org.opensearch.search.relevance.configuration.ResultTransformerConfiguration;
+import org.opensearch.search.relevance.configuration.ResultTransformerConfigurationFactory;
+import org.opensearch.search.relevance.transformer.kendraintelligentranking.KendraIntelligentRanker;
+
+import java.io.IOException;
+
+public class KendraIntelligentRankingConfigurationFactory implements ResultTransformerConfigurationFactory {
+    private KendraIntelligentRankingConfigurationFactory() {
+
+    }
```
noCharger: nit
```
  private KendraIntelligentRankingConfigurationFactory() {}
```

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/configuration/KendraIntelligentRankingConfiguration.java
```diff
@@ -39,7 +39,7 @@ public class KendraIntelligentRankingConfiguration extends ResultTransformerConf
     PARSER = new ObjectParser<KendraIntelligentRankingConfiguration, Void>("kendra_intelligent_ranking_configuration", KendraIntelligentRankingConfiguration::new);
     PARSER.declareInt(TransformerConfiguration::setOrder, TRANSFORMER_ORDER);
     PARSER.declareObject(KendraIntelligentRankingConfiguration::setProperties,
-        (p, c) -> KendraIntelligentRankingProperties.parse(p, c),
+            KendraIntelligentRankingProperties::parse,
```
noCharger, line 42: There are several of refactor from method reference operator (::) in this PR. I am wondering if we should choose one direction to make the style more consistent

```
        new SearchExtSpec<>(SearchConfigurationExtBuilder.NAME, SearchConfigurationExtBuilder::new, SearchConfigurationExtBuilder::parse));
        new SearchExtSpec<>(SearchConfigurationExtBuilder.NAME,
                input -> new SearchConfigurationExtBuilder(input, resultTransformerMap),
                parser -> SearchConfigurationExtBuilder.parse(parser, resultTransformerMap)));
```

noCharger (commented):
I guess `setting` class is more specific towards that transformer itself, whereas the `configuration` is the schema of the search workflow. There are two things I am not sure if we should put in this PR:
1. Can I use multiple same reranker, with different setting and configuration?
2. Do we want to provide more flexibility, such as one transformer depend on several other transformers.

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/client/KendraHttpClient.java
msfroh: > I guess `setting` class is more specific towards that transformer itself, whereas the `configuration` is the schema of the search workflow. There are two things I am not sure if we should put in this PR:
> 
> 1. Can I use multiple same reranker, with different setting and configuration?
> 2. Do we want to provide more flexibility, such as one transformer depend on several other transformers.

1. In the long run, that's the intention. I think it's out of the scope of this specific refactoring.
2. Transformers are not really intended to depend on each other. They get chained so one ends up impacting what the next one sees. The assumption, though, is that a transformer should be unaware of its neighbors.

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/client/KendraHttpClient.java
msfroh: Sure thing! I'll make those changes.

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/KendraIntelligentRanker.java
msfroh: ~~Right after which function signature?~~

Moved it up to right after the class signature

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/KendraIntelligentRanker.java
msfroh: Oh... it's an auto-fix suggested by IntelliJ (and I was able to apply it with Alt-Enter). 

`toArray` will allocate the array itself, and there's no real cost in allocating + passing an empty array.

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/configuration/KendraIntelligentRankingConfigurationFactory.java
msfroh: I'll remove the blank line, but I think our code style recommends not putting open/close on a single line, right?

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/configuration/KendraIntelligentRankingConfiguration.java
msfroh: IntelliJ recommends using the `::` operator where possible, so I usually autofix to that when it gets highlighted.

I'll double-check the rest of the PR to see if there are any that I missed.

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
msfroh: Will fix, thanks!

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
msfroh: Will fix, thanks!

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/client/KendraHttpClient.java
msfroh: Regarding the AccessController, this is a problem with the plugin model that we inherited, since it relies heavily on SecurityManager, which is also marked for deprecation. 

AFAIK, the fix will come with the broader OpenSearch move to extensions, which will use a more robust isolation model (while things that do provide core functionality will be moved into core).

src/main/java/org/opensearch/search/relevance/actionfilter/SearchActionFilter.java
msfroh: With this change, we are now logging the time that each individual transformer spent in the `transform` method, which would let us track down a single misbehaving transformer.

With the removal of the highlighted lines, we're no longer logging the time spent on everything else (e.g. loading index settings, the `shouldTransform` calls, the `preprocessRequest` calls).  I think that's okay for now. 

Note that convention suggests that we shouldn't log very much -- existing OpenSearch logs are not very noisy.

The correct solution, IMO, will be to return detailed tracing in the SearchResponse, if requested in the SearchRequest. (Maybe via the `stats` parameter?)

macohen (approved):

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/client/KendraHttpClient.java
```diff
@@ -50,45 +51,53 @@ public class KendraHttpClient {
...
-    amazonHttpClient = AccessController.doPrivileged((PrivilegedAction<AmazonHttpClient>) () -> new AmazonHttpClient(new ClientConfiguration()));
-    errorHandler = new SimpleAwsErrorHandler();
-    responseHandler = new SimpleResponseHandler();
-    aws4Signer = new AWS4Signer();
-    aws4Signer.setServiceName(KENDRA_RANKING_SERVICE_NAME);
-    aws4Signer.setRegionName(clientSettings.getServiceRegion());
-    objectMapper = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
     serviceEndpoint = clientSettings.getServiceEndpoint();
     executionPlanId = clientSettings.getExecutionPlanId();
+    if (isValid()) {
```
noCharger, line 60: Non-blocking:

This is what I preferred:

```
    if (!isValid()) {
      amazonHttpClient = null;
      aws4Signer = null;
      awsCredentialsProvider = null;
      errorHandler = null;
      responseHandler = null;
      return;
    }
    amazonHttpClient = AccessController.doPrivileged((PrivilegedAction<AmazonHttpClient>) () -> new AmazonHttpClient(new ClientConfiguration()));
    errorHandler = new SimpleAwsErrorHandler();
    responseHandler = new SimpleResponseHandler();
    aws4Signer = new AWS4Signer();
    aws4Signer.setServiceName(KENDRA_RANKING_SERVICE_NAME);
    aws4Signer.setRegionName(clientSettings.getServiceRegion());
   ...
```

noCharger (approved):
This is awesome! Approved with non-blocking comments.

macohen:
We do want to review the full set of metrics and logging events to capture. We'll create another issue for that purpose.

PR #58, opened 2022-12-21
Updating README.md to talk more about Search Query & Request Transfor…

macohen:
Signed-off-by: Mark Cohen <markcoh@amazon.com>

### Description
Updating to reflect the future of this repository better.
 
### Issues Resolved
N/A
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-[![Build and Test Search Reranker](https://github.com/opensearch-project/search-processor/actions/workflows/CI.yml/badge.svg)](https://github.com/opensearch-project/search-processor/actions/workflows/CI.yml)
+[![Build and Test Search Query & Request Transformers](https://github.com/opensearch-project/search-processor/actions/workflows/CI.yml/badge.svg)](https://github.com/opensearch-project/search-processor/actions/workflows/CI.yml)
 [![codecov](https://codecov.io/gh/opensearch-project/search-processor/branch/main/graph/badge.svg?token=PYQO2GW39S)](https://codecov.io/gh/opensearch-project/search-processor)
 ![PRs welcome!](https://img.shields.io/badge/PRs-welcome!-success)
 
-# OpenSearch Search Relevance - Reranker
+# Search Query & Request Transformers
 - [Welcome!](#welcome)
 - [Project Resources](#project-resources)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,7 +11,9 @@
 - [Copyright](#copyright)
 
 ## Welcome!
-This repository currently hosts a plugin to rerank search results before returning them to the client inline. The current plan is to expand this functionality to allow users to add rewriters, rerankers, and log what those components do. 
+This repository is the home of an evolving project to create a pipeline of transformers to preprocess search queries before information retrieval and post-process results after information retrieval. The first component here is hosts a plugin to re-rank search results before returning them to the client inline. In the coming year, we will add hooks to configure other re-rankers, and allow users to add their own components to the pipeline. Logging will also be a critical part of the pipeline in two ways: 1) Logging information about the search experience (e.g. query, search results returned from the index, search results returned to the OpenSearch client); 2) Logging debug information about the transformers.
```
noCharger: I'm not sure if `information retrieval` is the correct word to use instead of `search`. It reminds me the classic ways of search like https://en.wikipedia.org/wiki/Information_retrieval, even though we want to provide the capabilities of NLP and other ML techniques. @kolchfa-aws I would like to get Fanit's input on our knowledge sharing documents.

kolchfa-aws: I think it's better to be plain and precise, so I would use "search" instead of "information retrieval". The second sentence has "is hosts" which I'm guessing should be just "is"? Also, is it apparent what "component" means in this context?
```suggestion
This repository is the home of an evolving project that aims to create a pipeline of transformers to preprocess queries before search and post-process results after search. The first component here is a plugin to re-rank search results before returning them to the client inline. In the coming year, we will add hooks to configure other re-rankers and allow users to add their own components to the pipeline. Logging will also be a critical part of the pipeline in two ways: 
1. Logging information about the search experience (e.g. query, search results returned from the index, search results returned to the OpenSearch client) 
1. Logging debug information about the transformers
```

macohen: I was thinking of information retrieval as the step where we pull results from the actual index. Search is the whole pipeline, but for clarity I can accept this. Thanks.

noCharger (approved):
LGTM, thanks!

PR #103, opened 2023-02-14
Added untriaged issue workflow.

dblock:
Coming from https://github.com/opensearch-project/.github/issues/133, create a GitHub workflow for labelling untriaged issues. Please ensure that an untriaged label with color  exists before merging.


macohen:
We already have https://github.com/opensearch-project/search-processor/blob/main/.github/workflows/add-untriaged.yml which looks the same as the PR submitted here. Is there a difference I'm not seeing or can we close?

macohen (commented):
I don't see a difference between what's already in the repo and this PR. Am I missing something?

macohen:
@dblock Please reopen the PR if I missed a difference between what we already have and this PR. Thanks!

dblock:
Yeah, the PR was auto-generated and I think we have CR/LF differences.

PR #124, opened 2023-03-30
Updating build.gradle to use snapshot by default

sejli:
### Description
(Same as #123)
Updates `build.gradle` to follow other plugin patterns (in particular, using `System.getProperty` for using `SNAPSHOT` dependencies.
 
### Check List
  - [ ] All tests pass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,12 +53,11 @@ validateNebulaPom.enabled = false
 
 buildscript {
     ext {
-        plugin_version = "2.7.0"
-        snapshot = true // Set to false on release
-        if (snapshot) {
-            opensearch_version = plugin_version + "-SNAPSHOT"
-        } else {
-            opensearch_version = plugin_version
+        isSnapshot = "true" == System.getProperty("build.snapshot", "true")
+        opensearch_version = System.getProperty("opensearch.version", "2.7.0")
+        plugin_version = opensearch_version
+        if (isSnapshot) {
+            opensearch_version += "-SNAPSHOT"
         }
     }
 


msfroh (approved):

build.gradle
```diff
@@ -53,12 +53,11 @@ validateNebulaPom.enabled = false
...
 buildscript {
     ext {
-        plugin_version = "2.7.0"
-        snapshot = true // Set to false on release
-        if (snapshot) {
-            opensearch_version = plugin_version + "-SNAPSHOT"
-        } else {
-            opensearch_version = plugin_version
+        isSnapshot = "true" == System.getProperty("build.snapshot", "true")
+        opensearch_version = System.getProperty("opensearch.version", "2.7.0")
```
macohen, line 57: do we always want to have 2.7.0 as a default or should this blow up if opensearch.version is not available? seems like we could risk punting a failure down the road.

macohen (changes requested):

build.gradle
sejli: `System.getProperty` allows `build.gradle` to get values passed in from command line parameters. `2.7.0` would be the default value if nothing is passed in. For example, `./gradlew build -Dopensearch.version=2.6.0` would specify the version to be `2.6.0`. Likewise, `./gradlew build -Dbuild.snapshot=false` would specify that the build should be completed without snapshot dependencies.

mingshl (approved):

build.gradle
noCharger: According to my comprehension, this version should trace the most recent version and increment for each releasee.

macohen: In a scenario where we are on v6.2.1 (one day, maybe) and someone forgets to pass in 6.2.1 into the script, what happens? does it build 2.7.0? what happens then?

sejli: It's my understanding that we update this value per version upgrade. So, if we were working on `6.2.1`, then, on the `6.2` branch, we would have `System.getProperty("opensearch.version", "6.2.1")` (similarly, on the `6.x` branch, it would be `System.getProperty("opensearch.version", "6.3.0")` and on the `main` branch it would be `System.getProperty("opensearch.version", "7.0.0")`. In general, this value needs to be maintained per release and updated when necessary.

macohen (approved):

Issue #133, opened 2023-04-28
[Build] Jackson-core as a dependency


noCharger:
# Context

https://github.com/opensearch-project/search-processor/pull/132

Some GitHub action backend code using jackson-core 2.5.0 is failing builds because it includes an MRJAR with JDK 19 code. See, for example, [this failed action](https://github.com/opensearch-project/search-processor/actions/runs/4803442236/jobs/8547924201#step:5:27).

# Action items

- [x] For the short term, upgrade to jackson-core 2.5.0 in main branch to resolve the build failure. Also upgrade gradle https://github.com/opensearch-project/search-processor/pull/132 - Resolved here https://github.com/opensearch-project/search-processor/pull/134
- [ ] For the long term, we must devise a sustainable decision on `Jackson-core` as a dependency ref to Dan's suggestions [here](https://github.com/opensearch-project/search-processor/pull/132#issuecomment-1527733231). Can we remove all `jackson-` packages? if not. we need test on build and integration. [Non urgent]
- [ ] Check if the current [dependabot](https://github.com/opensearch-project/search-processor/blob/main/.github/workflows/dependabot_pr.yml) is working well, or examine alternatives.

Some observations:
1. OpenSearch main branch is using jackson-core 2.5.0 https://github.com/opensearch-project/OpenSearch/tree/main/client/sniffer/licenses
2. Security plugin removed jackson-databind and jackson-annotations https://github.com/opensearch-project/security/pull/2325
5. Observability plugin removed jackson depenencies and then added them back https://github.com/opensearch-project/observability/pull/1379

sejli:
Since we currently have build failures, we are opting to upgrade `jackson-core` to `2.5.0` for now, as implemented in #134. Alternatives are continued to be looked at.

macohen:
What's the latest on this? Should this be closed and a new issue created for long term?

PR #134, opened 2023-05-02
Fix build

msfroh:
Build was broken due to changes in OpenSearch core project that we needed to handle.

These changes include:
1. Gradle upgraded to 8.1
2. ImmutableOpenMap removed from OpenSearch core.
3. Jackson libs updated to 2.15.0


sejli (approved):

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,17 @@
...
+# compatible open source license.
+#
+# Modifications Copyright OpenSearch Contributors. See
+# GitHub history for details.
+#
+
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.1-all.zip
```
noCharger: From https://github.com/opensearch-project/search-processor/issues/133 "There are other incompatibilities in the build for Gradle 8, so 7.6.1 is the only option to keep a working build."

msfroh: Given that the build passes on this PR, what are the other incompatibilities? The linked issue doesn't explain why we can't use 8.1, like OpenSearch core.

I will try removing the Jackson dependencies altogether though. If we can just pick them up transitively, that should save us time in the long run.

msfroh: Hmm.... nope. I was able to remove jackson-core and have it compile, but if I remove either of jackson-databind or jackson-annotations, the build fails.

PR #135, opened 2023-05-02
Fetch index settings asynchronously

msfroh:
In some cases, like using ./gradlew run, I've run into issues where OpenSearch crashes, complaining that we're making a blocking call on a listener thread, because we fetch index settings to see if a result transformer has been configured for the current index.

I'm kind of surprised that we haven't run into this in production use-cases, but it may just be because assertions are not enabled in production. Regardless, blocking calls on listener threads are a bad idea, since that's how you run out of listener threads.

This change makes the index settings call asynchronous and chains the remaining logic off of that.


sejli (approved):
LGTM! I did notice that some of the formatting was changed on some files, particularly `ConfigurationUtils.java`. This repo should probably include the [Spotless Gradle plugin](https://github.com/diffplug/spotless/tree/main/plugin-gradle) for a formatting check.

PR #137, opened 2023-05-12
Add KendraRankingResponseProcessor

mingshl:
### Description
Add a KendraRankingResponseProcessor to enable creating a pipeline using `kendra-ranking`:

In this PR, users can create a pipeline with `kendra-ranking` response processor, and use the pipeline for searching:

sample command:

1. uploading documents:
```
curl -XPUT localhost:9200/test1/_doc/1 -d '{
  "message": "[123:1:0309/123054.737343412:ERROR:child_process_sandbox_support_impl_linux.cc(23)] FontService unique font name matching request did not receive a response.",
  "fileset": {
    "name": "syslog"
  },
  "process": {
    "name": "org.gnome.Shell.desktop",
    "pid": 1544
  },
  "@timestamp": "2020-03-09T18:00:54.000+05:30",
  "host": {
    "hostname": "bionic",
    "name": "bionic"
  }
}'  -H "Content-Type:Application/json"
```

2. creating a new pipeline named `test-kendra`
```
curl -XPUT 'localhost:9200/_search/pipeline/test-kendra' --data '
{
  "description": "A pipeline to apply kendra ranking",
  "response_processors" : [
    {
      "kendra_ranking": {
        "title_field": "fileset",
        "body_field": "message",
        "doc_limit": 500
      }
    }
  ]
}' -H 'Content-Type: application/json' 
```

3. searching with the new pipeline `test-kendra` with normal query, for example match query: 

```
curl -XGET 'localhost:9200/test1/_search?search_pipeline=test-kendra' --data '{
  "query": {
    "match": {
      "body_field": "2020-03-09T18:00:54.000+05:30"
    }
  }
' -H 'Content-Type: application/json' 

```

4. Expecting a response with re-rank outcomes.

 
### Issues Resolved
[#7158](https://github.com/opensearch-project/OpenSearch/issues/7158)
 
### Check List
  - [ ] All tests pass
- [ ] Commits are signed as per the DCO using --signoff 

### DIFF
--- a/src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/pipeline/KendraRankingResponseProcessor.java
+++ b/src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/pipeline/KendraRankingResponseProcessor.java
@@ -1,0 +1,173 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.search.relevance.transformer.kendraintelligentranking.pipeline;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.action.search.SearchResponse;
+import org.opensearch.action.search.SearchResponseSections;
+import org.opensearch.ingest.ConfigurationUtils;
+import org.opensearch.search.SearchHits;
+import org.opensearch.search.aggregations.InternalAggregations;
+import org.opensearch.search.internal.InternalSearchResponse;
+import org.opensearch.search.pipeline.Processor;
+import org.opensearch.search.pipeline.SearchResponseProcessor;
+import org.opensearch.search.profile.SearchProfileShardResults;
+import org.opensearch.search.relevance.transformer.kendraintelligentranking.KendraIntelligentRanker;
+import org.opensearch.search.relevance.transformer.kendraintelligentranking.client.KendraClientSettings;
+import org.opensearch.search.relevance.transformer.kendraintelligentranking.client.KendraHttpClient;
+import org.opensearch.search.relevance.transformer.kendraintelligentranking.configuration.KendraIntelligentRankingConfiguration;
+
+import static org.opensearch.search.relevance.transformer.kendraintelligentranking.configuration.Constants.KENDRA_DEFAULT_DOC_LIMIT;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/pipeline/KendraRankingResponseProcessor.java
```diff
@@ -0,0 +1,174 @@
...
+    private static final Logger logger = LogManager.getLogger(KendraRankingResponseProcessor.class);
+
+    /**
+     * Constructor that apply configuration for kendra re-ranking
+     *
+     * @param tag          processor tag
+     * @param description  processor description
+     * @param titleField   titleField applied to kendra re-ranking
+     * @param bodyField    bodyField applied to kendra re-ranking
+     * @param docLimit     docLimit applied to kendra re-ranking
```
msfroh: Looks like the build is complaining about the Javadoc for this param. Should be `@param inputDocLimit`.

src/test/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/pipeline/KendraRankingResponseProcessorTests.java
```diff
@@ -0,0 +1,112 @@
...
+        shortConfiguration.put("body_field","body");
+        KendraRankingResponseProcessor processorWithOneFields = factory.create(Collections.emptyMap(),"tmp1","testingBodyField", shortConfiguration);
+        assertEquals(TYPE, processorWithOneFields.getType());
+        assertEquals("tmp1", processorWithOneFields.getTag());
+        assertEquals("testingBodyField", processorWithOneFields.getDescription());
+
+
+    }
+    public void testRankingResponse() throws Exception {
+        KendraHttpClient kendraClient = new KendraHttpClient(KendraClientSettings.getClientSettings(env.settings()));
```
msfroh: Does this run as a unit test? It looks like it's creating a real KendraHttpClient.

Oh... or maybe it runs but doesn't call Kendra Intelligent Ranking because the client is not valid, so the whole `isValid()` check skips it. In this case, the responses will get returned unmodified, right?

mingshl: Right. It should create a Client, but because there is not valid endpoint and credentials, so it shows warning that "it's skipping kendra client and return response without re-ranking"

src/main/java/org/opensearch/search/relevance/transformer/kendraintelligentranking/pipeline/KendraRankingResponseProcessor.java
mingshl: fixed docstring

msfroh (approved):

PR #138, opened 2023-05-18
Personalized intelligent ranking for open search requests

kulket:
Description:
============
This change uses search pipeline feature of open search to allow customers to create a response processor.
Response processor will rerank search results obtained from open search using Personalization solution before returning results to the user.

Check List:
==========
  - [x] All tests pass
- [x] New functionality has been documented.

Signed-off-by: Ketan Kulkarni <kektnr@amazon.com>

### DIFF
--- a/src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeRankingResponseProcessor.java
+++ b/src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeRankingResponseProcessor.java
@@ -33,13 +33,6 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
 
-import static org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.Constants.CAMPAIGN_ARN_CONFIG_NAME;
-import static org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.Constants.IAM_ROLE_ARN_CONFIG_NAME;
-import static org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.Constants.RECIPE_CONFIG_NAME;
-import static org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.Constants.REGION_CONFIG_NAME;
-import static org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.Constants.ITEM_ID_FIELD_CONFIG_NAME;
-import static org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.Constants.WEIGHT_CONFIG_NAME;
-
 /**
  * This is a {@link SearchResponseProcessor} that applies Personalized intelligent ranking
  */
@@ -138,6 +131,12 @@ public class PersonalizeRankingResponseProcessor implements SearchResponseProces
 
     public static final class Factory implements Processor.Factory<SearchResponseProcessor> {
 
+        private static final String CAMPAIGN_ARN_CONFIG_NAME = "campaign_arn";
+        private static final String ITEM_ID_FIELD_CONFIG_NAME = "item_id_field";
+        private static final String IAM_ROLE_ARN_CONFIG_NAME = "iam_role_arn";
+        private static final String RECIPE_CONFIG_NAME = "recipe";
+        private static final String REGION_CONFIG_NAME = "aws_region";
+        private static final String WEIGHT_CONFIG_NAME = "weight";
         PersonalizeClientSettings personalizeClientSettings;
         private final BiFunction<AWSCredentialsProvider, String, PersonalizeClient> clientBuilder;
 
@@ -152,7 +151,6 @@ public class PersonalizeRankingResponseProcessor implements SearchResponseProces
 
         @Override
         public PersonalizeRankingResponseProcessor create(Map<String, Processor.Factory<SearchResponseProcessor>> processorFactories, String tag, String description, Map<String, Object> config) throws Exception {
-            // TODO: Handle validation as well as required vs non required config differentiation with related user error / exception
             String personalizeCampaign = ConfigurationUtils.readStringProperty(TYPE, tag, config, CAMPAIGN_ARN_CONFIG_NAME);
             String iamRoleArn = ConfigurationUtils.readOptionalStringProperty(TYPE, tag, config, IAM_ROLE_ARN_CONFIG_NAME);
             String recipe = ConfigurationUtils.readStringProperty(TYPE, tag, config, RECIPE_CONFIG_NAME);
--- a/src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeRankingResponseProcessor.java
+++ b/src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeRankingResponseProcessor.java
@@ -31,6 +31,7 @@ import org.opensearch.search.relevance.transformer.personalizeintelligentranking
 
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiFunction;
 
 import static org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.Constants.CAMPAIGN_ARN_CONFIG_NAME;
 import static org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.Constants.IAM_ROLE_ARN_CONFIG_NAME;
@@ -138,9 +139,15 @@ public class PersonalizeRankingResponseProcessor implements SearchResponseProces
     public static final class Factory implements Processor.Factory<SearchResponseProcessor> {
 
         PersonalizeClientSettings personalizeClientSettings;
+        private final BiFunction<AWSCredentialsProvider, String, PersonalizeClient> clientBuilder;
+
+        Factory(PersonalizeClientSettings settings, BiFunction<AWSCredentialsProvider, String, PersonalizeClient> clientBuilder) {
+            this.personalizeClientSettings = settings;
+            this.clientBuilder = clientBuilder;
+        }
 
         public Factory(PersonalizeClientSettings settings) {
-            this.personalizeClientSettings = settings;
+            this(settings, PersonalizeClient::new);
         }
 
         @Override
@@ -155,9 +162,8 @@ public class PersonalizeRankingResponseProcessor implements SearchResponseProces
 
             PersonalizeIntelligentRankerConfiguration rankerConfig =
                     new PersonalizeIntelligentRankerConfiguration(personalizeCampaign, iamRoleArn, recipe, itemIdField, awsRegion, weight);
-            PersonalizeCredentialsProviderFactory credentialsProviderFactory = new PersonalizeCredentialsProviderFactory();
-            AWSCredentialsProvider credentialsProvider = credentialsProviderFactory.getCredentialsProvider(personalizeClientSettings, iamRoleArn, awsRegion);
-            PersonalizeClient personalizeClient = new PersonalizeClient(credentialsProvider, awsRegion);
+            AWSCredentialsProvider credentialsProvider = PersonalizeCredentialsProviderFactory.getCredentialsProvider(personalizeClientSettings, iamRoleArn, awsRegion);
+            PersonalizeClient personalizeClient = clientBuilder.apply(credentialsProvider, awsRegion);
             return new PersonalizeRankingResponseProcessor(tag, description, rankerConfig, personalizeClient);
         }
     }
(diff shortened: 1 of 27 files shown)


src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeRankingResponseProcessor.java
```diff
@@ -0,0 +1,160 @@
...
+
+        @Override
+        public PersonalizeRankingResponseProcessor create(Map<String, Processor.Factory> processorFactories, String tag, String description, Map<String, Object> config) throws Exception {
+            // TODO: Handle validation as well as required vs non required config differentiation with related user error / exception
+            String personalizeCampaign = ConfigurationUtils.readStringProperty(TYPE, tag, config, CAMPAIGN_ARN_CONFIG_NAME);
+            String iamRoleArn = ConfigurationUtils.readStringProperty(TYPE, tag, config, IAM_ROLE_ARN_CONFIG_NAME);
+            String recipe = ConfigurationUtils.readStringProperty(TYPE, tag, config, RECIPE_CONFIG_NAME);
+            String itemIdField = ConfigurationUtils.readStringProperty(TYPE, tag, config, ITEM_ID_FIELD_CONFIG_NAME);
+            String awsRegion = ConfigurationUtils.readStringProperty(TYPE, tag, config, REGION_CONFIG_NAME);
+            double weight = Double.parseDouble(ConfigurationUtils.readStringProperty(TYPE, tag, config, WEIGHT_CONFIG_NAME));
```
msfroh: There's a `ConfigurationUtils.readDoubleProperty` that you could use here.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/client/PersonalizeClientSettings.java
```diff
@@ -0,0 +1,95 @@
...
+        try (SecureString key = ACCESS_KEY_SETTING.get(settings);
+             SecureString secret = SECRET_KEY_SETTING.get(settings);
+             SecureString sessionToken = SESSION_TOKEN_SETTING.get(settings)) {
+            if (key.length() == 0 && secret.length() == 0) {
+                if (sessionToken.length() > 0) {
+                    throw new SettingsException("Setting [{}] is set but [{}] and [{}] are not",
+                            SESSION_TOKEN_SETTING.getKey(), ACCESS_KEY_SETTING.getKey(), SECRET_KEY_SETTING.getKey());
+                }
+                logger.info("Using either environment variables, system properties or instance profile credentials");
+                return null;
```
msfroh, line 68: Should this return a `DefaultAWSCredentialsProviderChain`? 

I guess you would need the other branch to return an instance of AWSCredentialsProvider too, but you could just wrap the return value in an instance of `AWSStaticCredentialsProvider`.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeRankingResponseProcessor.java
kulket: Updated.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/client/PersonalizeClientSettings.java
kulket: This method is used for getting credentials from open search keystore. There is another class PersonalizeCredentailsProviderFactory which implements logic for providing appropriate implementation of AWSCredentialsProvider based on availability of different credentials options.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/client/PersonalizeCredentialsProviderFactory.java
```diff
@@ -0,0 +1,87 @@
...
+    private static final Logger logger = LogManager.getLogger(PersonalizeCredentialsProviderFactory.class);
+    private static final String ASSUME_ROLE_SESSION_NAME = "OpenSearchPersonalizeIntelligentRankingPluginSession";
+
+    /**
+     * Get AWS credentials provider either from static credentials from open search keystore or
+     * using DefaultAWSCredentialsProviderChain.
+     * @param clientSettings Personalize client settings
+     * @return AWS credentials provider for accessing Personalize
+     */
+    public AWSCredentialsProvider getCredentialsProvider(PersonalizeClientSettings clientSettings) {
```
msfroh: Does this one get called from anywhere else? Can it be `private` + `static`?

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/client/PersonalizeCredentialsProviderFactory.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Get AWS credentials provider by assuming IAM role if provided or else
+     * use static credentials or DefaultAWSCredentialsProviderChain.
+     * @param clientSettings        Personalize client settings
+     * @param personalizeIAMRole    IAM role configuration for accessing Personalize
+     * @param awsRegion             AWS region
+     * @return AWS credentials provider for accessing Amazon Personalize
+     */
+    public AWSCredentialsProvider getCredentialsProvider(PersonalizeClientSettings clientSettings,
```
msfroh: This method can be `static`.

There's no value in creating an instance of this class, since it's just a helper method to handle credential providers.

msfroh:
@kulket -- Can you please rebase from main and force-push to your fork? 

If you check the list of commits associated with this PR, it includes the Jackson-core update, which isn't actually part of this PR.

kulket:
> @kulket -- Can you please rebase from main and force-push to your fork?
> 
> If you check the list of commits associated with this PR, it includes the Jackson-core update, which isn't actually part of this PR.

Performed rebased as well as force push.
I am seeing 3 commits now.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeRankingResponseProcessor.java
```diff
@@ -0,0 +1,170 @@
...
+import org.opensearch.search.relevance.transformer.personalizeintelligentranking.requestparameter.PersonalizeRequestParameterUtil;
+import org.opensearch.search.relevance.transformer.personalizeintelligentranking.requestparameter.PersonalizeRequestParameters;
+import org.opensearch.search.relevance.transformer.personalizeintelligentranking.reranker.PersonalizedRanker;
+import org.opensearch.search.relevance.transformer.personalizeintelligentranking.reranker.PersonalizedRankerFactory;
+
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.function.BiFunction;
+
+import static org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.Constants.CAMPAIGN_ARN_CONFIG_NAME;
```
mingshl: it seems that the class is importing individual configuration constants from the configuration package. While this approach can work, I believe it might be more maintainable and flexible to incorporate the configuration as an attribute of the plugins and access it using `this.personalizeClientsetting`. This way, the class can receive the configuration as a parameter, allowing for greater modularity and easier testing.

Check out this PR, which use `this.kendraClientsetting`
https://github.com/opensearch-project/search-processor/blob/01ca2045eca5bc1c5ebba7a552e362738c1868be/src/main/java/org/opensearch/search/relevance/SearchRelevancePlugin.java#L95

https://github.com/opensearch-project/search-processor/blob/01ca2045eca5bc1c5ebba7a552e362738c1868be/src/main/java/org/opensearch/search/relevance/SearchRelevancePlugin.java#L119

kulket: Hi mingshl, 
Thank you for your review and comment.

I kept code for getting ``PersonalizeClientSettings`` instance to limited to the ``PersonalizeRankingResponseProcessor.Factory`` instance creation so that ``PersonalizeClientSettings`` are not created when they are not required. For e.g. Creating that instance outside factory creation may create  ``PersonalizeClientSettings`` instance when using plugin for Kendra only use case.

I did thought about creating one common class for Kendra and Personalize related settings. But Kendra needs more configurations than just credentials and hence opted to keep separate class for Personalize.

Let me know your thoughts regarding above.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/client/PersonalizeClient.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param request Get personalized ranking request
+     * @return Personalized ranking results
+     */
+    public GetPersonalizedRankingResult getPersonalizedRanking(GetPersonalizedRankingRequest request) {
+        GetPersonalizedRankingResult result;
+        try {
+            result = AccessController.doPrivileged(
+                    (PrivilegedAction<GetPersonalizedRankingResult>) () -> personalizeRuntime.getPersonalizedRanking(request));
+        } catch (AmazonServiceException ex) {
+            throw ex;
```
noCharger, line 60: Should this be a direct rethrow or an OpenSearchException?

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeRankingResponseProcessor.java
msfroh: I guess another question is whether it makes sense to declare the `Factory` parameters in a separate `Constants` class just to use them here. They're *only* used to pass values to the processor `Factory`. As such, they can be kept in this class.

A "global" `Constants` class is only really useful for values used in multiple classes. Otherwise, it's just an extra hop to find values that are only used in one class.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/reranker/impl/AmazonPersonalizedRankerImpl.java
```diff
@@ -0,0 +1,106 @@
...
+            // TODO: Parse context from request parameters
+            String userId = requestParameters.getUserId();
+            logger.info("User ID from request parameters. User ID: {}", userId);
+            GetPersonalizedRankingRequest personalizeRequest = new GetPersonalizedRankingRequest()
+                    .withCampaignArn(rankerConfig.getPersonalizeCampaign())
+                    .withInputList(documentIdsToRank)
+                    .withUserId(userId);
+            GetPersonalizedRankingResult result = personalizeClient.getPersonalizedRanking(personalizeRequest);
+
+            //TODO: Combine Personalize and open search result. Change the result after transform logic is implemented
```
noCharger, line 78: Seems to unfinished impl. If it's impl else where, please remove the TODOs

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/reranker/PersonalizedRankerFactory.java
```diff
@@ -0,0 +1,40 @@
...
+import org.opensearch.search.relevance.transformer.personalizeintelligentranking.client.PersonalizeClient;
+import org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.PersonalizeIntelligentRankerConfiguration;
+import org.opensearch.search.relevance.transformer.personalizeintelligentranking.reranker.impl.AmazonPersonalizedRankerImpl;
+
+import static org.opensearch.search.relevance.transformer.personalizeintelligentranking.configuration.Constants.AMAZON_PERSONALIZED_RANKING_RECIPE_NAME;
+
+/**
+ * Factory for creating Personalize ranker instance based on Personalize ranker configuration
+ */
+public class PersonalizedRankerFactory {
```
noCharger, line 21: If there is only one impl expected, can we simplify the code structure without utilizing a factory?

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeRankingResponseProcessor.java
kulket: Makes sense. Updated code to move search pipeline related configuration constants to Factory class instead of using earlier approach of global constants.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/reranker/PersonalizedRankerFactory.java
kulket: We expect more than one implementation going forward so use factory approach here.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/reranker/impl/AmazonPersonalizedRankerImpl.java
kulket: We will update unfinished implementation as a follow up PR.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/client/PersonalizeClient.java
kulket: Will be using follow up PR to update exception handling mentioned here.

mingshl (approved):
LGTM

PR #144, opened 2023-05-26
Support contextual metadata to use when getting personalized reranking

kulket:
This change supports using contextual metadata for personalized reranking if provided as a search request parameter

### Description
This change supports using contextual metadata for personalized reranking if provided as a search request parameter
 
### Check List
  - [X] All tests pass

Signed-off-by: Ketan Kulkarni [kektnr@amazon.com](mailto:kektnr@amazon.com)


src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/reranker/impl/AmazonPersonalizedRankerImpl.java
```diff
@@ -66,12 +65,21 @@ public SearchHits rerank(SearchHits hits, PersonalizeRequestParameters requestPa
                         .collect(Collectors.toList());
             }
             logger.info("Document Ids to re-rank with Personalize: {}", Arrays.toString(documentIdsToRank.toArray()));
-            // TODO: Parse context from request parameters
             String userId = requestParameters.getUserId();
+            Map<String, String> context = requestParameters.getContext() != null ?
+                                            requestParameters.getContext().entrySet().stream()
+                                                    .filter(e -> e.getValue() instanceof String)
```
msfroh: Should probably throw an exception for any non-string parameters, right? 

Or you could throw an exception on explicit `null` values and otherwise invoke `toString` on any parameters.

kulket: Agree. Personalize API expects Map<String, String> for context input. So decided to throw exception instead of assuming toString() returns correct value.

src/main/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/reranker/impl/AmazonPersonalizedRankerImpl.java
```diff
@@ -66,12 +65,27 @@ public SearchHits rerank(SearchHits hits, PersonalizeRequestParameters requestPa
...
-            // TODO: Parse context from request parameters
             String userId = requestParameters.getUserId();
+            Map<String, String> context = requestParameters.getContext() != null ?
+                                            requestParameters.getContext().entrySet().stream()
+                                                    .filter(e -> {
+                                                        try {
+                                                            return isValidPersonalizeContext(e);
+                                                        } catch (IllegalArgumentException iae) {
+                                                            throw iae;
+                                                        }
```
msfroh: This catch + rethrow isn't a great pattern.

A better bet is to call the helper method something like `validateIsString` and make it:

```
    private static String validateIsString(Object obj) {
        if (obj instanceof String) {
            return (String) obj;
        } else {
            throw new IllegalArgumentException("Personalize context value is not of type String. " +
                    "Invalid context value: " + obj;
        }
    }

   // Then in the rerank method:

    Map<String, String> context = null;
    if (requestParameters.getContext() != null) {
        context = requestParameters.getContext().entrySet().stream()
                .collect(Collectors.toMap(Map.Entry::getKey, e -> validateIsString(e.getValue()));
    }

```

kulket: Sure. Removed catch + rethrow pattern with latest commit.
Thanks for suggestion.

src/test/java/org/opensearch/search/relevance/transformer/personalizeintelligentranking/PersonalizeResponseProcessorTests.java
```diff
@@ -131,4 +131,82 @@ public void testProcessorWithHits() throws Exception {
...
+        configuration.put("recipe", AMAZON_PERSONALIZED_RANKING_RECIPE_NAME);
+        configuration.put("weight", String.valueOf(weight));
+        configuration.put("iam_role_arn", iamRoleArn);
+        configuration.put("aws_region", region);
+
+        PersonalizeRankingResponseProcessor personalizeResponseProcessor =
+                factory.create(Collections.emptyMap(), "testTag", "testingAllFields", configuration);
+
+        Map<String, Object> personalizeContext = new HashMap<>();
+        personalizeContext.put("contextKey2", 5);
```
msfroh, line 192: Shouldn't this trigger an exception in the call to `processResponse`?

msfroh: Oh -- the exception gets caught within `processResponse`. Got it

msfroh (approved):

PR #177, opened 2023-07-21
Helper script for quickstart setup of Personalize plugin

kulket:
### Description:
Helper script is intended for quick start setup for OpenSearch. Script does following:

* Build containers for both ``opensearch`` and ``opensearch dashboard``
* Picks up credentials from given profile or environment variable or default profile and stores them in opensearch keystore if available.
* Gets both ``opensearch`` and ``opensearch dashboard`` up and running
 
### Check List
  - [X] All tests pass

### DIFF
--- a/helpers/personalized_search_ranking_quickstart.sh
+++ b/helpers/personalized_search_ranking_quickstart.sh
@@ -1,0 +1,387 @@
+#!/bin/bash
+
+set -o errexit
+set -o errtrace
+set -o pipefail
+set -o nounset
+
+# Some useful constants
+readonly DOCKER_IMAGE_TAG="opensearch-with-personalized-search-ranking"
+readonly OPENSEARCH_VERSION="2.9.0"
+
+#
+# Set default values for OpenSearch (+Dashboards) image tags and plugin URL.
+# We can pass them from outside to override these settings for testing purposes.
+#
+if [ -z "${OPENSEARCH_IMAGE_TAG:-}" ]; then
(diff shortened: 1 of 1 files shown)


helpers/personalized_search_ranking_quickstart.sh
```diff
@@ -0,0 +1,387 @@
+#!/bin/bash
+
+set -o errexit
+set -o errtrace
+set -o pipefail
+set -o nounset
+
+# Some useful constants
+readonly DOCKER_IMAGE_TAG="opensearch-with-personalized-search-ranking"
+readonly OPENSEARCH_VERSION="2.9.0"
```
mingshl, line 10: if hard coded the version, in the following version upgrade, you will need to bump the version here, and also you are not able to run the script in main with 3.0.0. 
 
 Suggested to read the version number from the build.gradle file [here](https://github.com/opensearch-project/search-processor/blob/e6fe9ed9808e0c7f8adddd0cd8562b2a25d2dfb1/build.gradle#L57)

kulket: After current release, I am actually planning to update ``OPENSEARCH_VERSION`` to use tag value ``latest`` instead of any specific version in order to always use latest publicly available version.

Since this script depends upon availability of container image I don't think havning future release value like ```3.0.0``` is feasible. 

Please let me know if I am missing something as well as if it is ok to update tag value to ``latest`` as a follow up change once v 2.9.0 is released.

mingshl: sounds good.

mingshl (approved):

PR #194, opened 2023-08-28
Update README to Reflect Approach to Processors

### LINKED ISSUE #129: [DOC] Update README File to Reflect What This Repo Contains
This one has been back and forth a few times, but now that Search Pipelines is in core, we should update the README to reflect that this contains the Kendra self-install plugin and will be a shim for the processor being built. Is this also the future of this repo long term?

macohen:
### Description
Acknowledges that this repository holds the Kendra and Personalize rerankers and points to user documentation for search pipelines and these processors.
 
### Issues Resolved
#129 
 
### Check List


README.md
```diff
@@ -2,29 +2,26 @@
...
-1. Logging debug information about the transformers
+This repository hosts the code for two self-install re-rankers that integrate into [Search Pipelines](https://opensearch.org/docs/latest/search-plugins/search-pipelines/index/). User documentation for the Personalize Reranker is [here](https://opensearch.org/docs/latest/search-plugins/search-pipelines/personalize-search-ranking/). For Kendra, it is [here](https://opensearch.org/docs/latest/search-plugins/search-relevance/index/#reranking-results-with-kendra-intelligent-ranking-for-opensearch). The current guideline for developing processors is that if you are developing a processor that makes a network connection outside of OpenSearch, it should be in a separate repository. If it does not make a separate connection, it could go into the OpenSearch repository under [org.opensearch.search.pipeline.common](https://github.com/opensearch-project/OpenSearch/tree/a08d588691c3b232e65d73b0a0c2fc5c72c870cf/modules/search-pipeline-common). Please create a PR there if you have an idea for a new processor.
```
noCharger: >The current guideline for developing processors is that if you are developing a processor that makes a network connection outside of OpenSearch, it should be in a separate repository

If someone wishes to add a new processor argument search results by obtaining data from a Spark cluster, it may be a broad data source processor or a very customized Spark processor implementation. Do they both fall within this category? Should we advise that the new processor reside in this repo or in a separate repo in the opensearch-project?

macohen: That's a good point. Can you expand on the difference between a "broad data source processor vs customized Spark processor?" I would not want to put any specific business logic in core. I wouldn't even want to put that in a separate repo. A Spark job definition (via SQL or a specific implementation in a JVM language (or I suppose Python)) should be passed to the processor as a parameter. 

We may look to the connectors framework under development to define this better. There is one implementation (Apache Spark) and at least a few deployments of Spark (e.g. Databricks, EMR). I'm not sure if the way you connect to each is different, but I think I would want to leave that detail out of a generic Spark processor. 

More concretely, the DynamoDB processor could be a good example. Because it's a specific database implementation I would want to see much of that outside of core, but there may be helpers or other ways to make it easier for developers to create any number of processors that pull data from a database. Are there similarities between Cassandra, DynamoDB, Oracle, MongoDB, MySQL, and even Spark when it comes to processors that we might factor out and put in core?

What I do believe is that no more processors belong here. Do we agree on that? We can always ask for PRs against core if people want to create a processor. Maintainers can then judge whether the processor belongs there or somewhere else.

cc: @msfroh

macohen: Rethinking this a bit: we should consider guidelines in terms of dependencies. Introducing new dependencies into @opensearch-project/OpenSearch should be discouraged. In the case of Spark, I would say that if there are abstractions that could make it easier to bring a processor in that access external data sources, we should consider those in core, but I would have a hard time adding a dependency on Spark in core for almost any reason. Will reflect this in the next revision.

noCharger (approved):

PR #214, opened 2023-12-15
Addresses changes made to security demo config install tool

DarshitChanpura:
### Description
There were recent changes to security demo configuration setup which now requires a custom admin Password to be setup. This PR updates the documentation and related code change.
 
### Check List
  - [ ] All tests pass


msfroh:
Thanks @DarshitChanpura! 

Before we merge this, I'm going to make some changes to update the quickstart scripts to actually use newer versions of OpenSearch. (Currently, the Kendra one still downloads OpenSearch 2.7 and the Personalize one downloads 2.9.)

derek-ho (approved):
LGTM

helpers/personalized_search_ranking_quickstart.sh
```diff
@@ -252,6 +252,13 @@ if [ -n "${VOLUME_NAME:-}" ]; then
     external: true"
 fi
 echo "Volume created"
+
+# OpenSearch 2.12 onwards security plugins requires a password to be set to setup admin user
+if [ "$(echo "${OPENSEARCH_VERSION} 2.12" | awk '{print ($1 >= $2)}')" -eq 1 ] && [ -z "${OPENSEARCH_INITIAL_ADMIN_PASSWORD}" ]; then
+  echo "OpenSearch 2.12 onwards, the Security Plugins requires initial admin password to be set for demo config setup"
+  exit 1
+fi
```
derek-ho: This doesn't work comparing 2.7 vs 2.12 - I suggest:
```
IFS='.' read -ra version_array <<< "$VERSION"



if [ -z "$CREDENTIAL" ]
then
  # Starting in 2.12.0, security demo configuration script requires an initial admin password
  if (( ${version_array[0]} > 2 || (${version_array[0]} == 2 && ${version_array[1]} >= 12) )); then
      CREDENTIAL="admin:myStrongPassword123!"
  else
      CREDENTIAL="admin:admin"
  fi
fi
```

derek-ho: Which is the logic I used in other repos

DarshitChanpura:
@msfroh @macohen @macohen @noCharger @mingshl @sejli Can one of the maintainers take a look at the failing CI and help bring this to closure?

helpers/personalized_search_ranking_quickstart.sh
```diff
@@ -252,6 +252,16 @@ if [ -n "${VOLUME_NAME:-}" ]; then
...
 fi
 echo "Volume created"
+
+# Starting in 2.12.0, security demo configuration script requires an initial admin password
+OPENSEARCH_REQUIRED_VERSION="2.12.0"
+COMPARE_VERSION=`echo $OPENSEARCH_REQUIRED_VERSION $OPENSEARCH_VERSION | tr ' ' '\n' | sort -V | uniq | head -n 1`
+if [ "$COMPARE_VERSION" != "$OPENSEARCH_REQUIRED_VERSION" ]; then
+  OPENSEARCH_INITIAL_ADMIN_PASSWORD="admin"
+else
+  OPENSEARCH_INITIAL_ADMIN_PASSWORD="myStrongPassword123!"
```
msfroh: Rather than hard-coding, we should probably take the admin password as a command-line arg.

@DarshitChanpura -- did earlier versions support the `OPENSEARCH_INITIAL_ADMIN_PASSWORD` env var? Or does that change come in 2.12? 

I'm wondering if it makes sense to change the scripts to require the initial admin password for all versions, where it will be required for 2.12 and a good idea for earlier versions.

DarshitChanpura: > Rather than hard-coding, we should probably take the admin password as a command-line arg.

Starting 2.12 a custom strong password is required. 

> @DarshitChanpura -- did earlier versions support the OPENSEARCH_INITIAL_ADMIN_PASSWORD env var? Or does that change come in 2.12?

This change is introduced in 2.12 and will be present for all versions 2.12 and above

> I'm wondering if it makes sense to change the scripts to require the initial admin password for all versions, where it will be required for 2.12 and a good idea for earlier versions.

That might be your design decision as a maintainer. Since the admin password is required only for 2.12 and above, IMO we can keep a version check in place and require it only for >= 2.12

msfroh: This doesn't answer my question. Allow me to rephrase:

Is it **POSSIBLE** to set the admin password in `docker-compose.yml` before 2.12?

If so, I will change the script to always require a custom password, for all versions, since it's a good idea.

DarshitChanpura: > Is it POSSIBLE to set the admin password in docker-compose.yml before 2.12?

You can set the env variable `OPENSEARCH_INITIAL_ADMIN_PASSWORD` in the docker-compose.yml. This value will only be picked up for 2.12 and above

You, can change the script to have a default strong password, and then based on the version it will be picked.

Does that answer your query?

msfroh: Yes, thank you.

In that case, it does make more sense to accept and enforce it for version 2.12 and higher only, because requiring the parameter on earlier versions and then ignoring it (leaving the default password as `admin`) would be a terrible user experience. 

In fact, we should reject the parameter (or at least output a warning that it's being ignored) if the version is pre-2.12.

I'll make those changes to the script.

msfroh: Made those changes and pushed the commit to this PR. 

@DarshitChanpura, please let me know what you think. Thanks!

DarshitChanpura (commented):
LGTM!

msfroh (approved):

noCharger (approved):
LGTM

msfroh:
> The backport to 2.12 failed:

That makes sense, since we haven't cut a 2.12 branch yet -- it will get cut from 2.x.

I'm just going to remove the `backport-2.12` label.

